PR #27, opened 2026-01-21
Update Homebrew tap and install command in README

dl-alexandre:


rudrankriyam:
Thanks @dl-alexandre for the update! I aligned the other Homebrew install section to the new tap so the README is consistent.

PR #419, opened 2026-02-03
feat(auth): add offline auth doctor diagnostics

mitre88:
Adds  (offline) to validate config/profile credentials and private key path. Includes JSON output for CI and unit tests.


mitre88:
Opened by mistake; intended for mitre88/App-Store-Connect-CLI. Closing.

PR #425, opened 2026-02-05
publish: resolve TestFlight groups by name or ID

rudrankriyam:
## Summary
- allow `asc publish testflight --group` to resolve group names as well as IDs
- paginate through all beta groups when resolving names
- return explicit errors for missing/ambiguous group names and dedupe resolved IDs
- add a release automation skill reference in `Agents.md`
- add a new skill at `skills/asc-release-successor/SKILL.md`

## Testing
- go test ./...

### DIFF
--- a/internal/cli/publish/group_resolver.go
+++ b/internal/cli/publish/group_resolver.go
@@ -3,6 +3,7 @@ package publish
 import (
 	"context"
 	"fmt"
+	"slices"
 	"strings"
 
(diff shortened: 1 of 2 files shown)


rudrankriyam:
Addressed one follow-up gap:
- added pagination-path coverage for beta group resolution in 
- refactored resolver helper to depend on a tiny interface so pagination behavior is testable without network

Validation:
- go test ./internal/cli/publish ./internal/cli/cmdtest
- go test ./...

PR #436, opened 2026-02-07
migrate: follow-up hardening for appInfo selection

rudrankriyam:
## Summary
Follow-up PR after #434 and #435 to harden `migrate` appInfo selection and remove drift between import/export paths.

## The 5 follow-up changes
1. **Unify selector usage**: `migrate export` now uses `selectBestAppInfoID` (same as `migrate import`) instead of a separate helper.
2. **Remove duplicate selector path**: deleted `pickEditableAppInfoID` to avoid divergent logic over time.
3. **Nil/empty safety**: `selectBestAppInfoID` now safely handles `nil` and empty responses.
4. **State fallback hardening**: fallback selection now considers both `state` and `appStoreState` and treats live states (`READY_FOR_DISTRIBUTION`, `READY_FOR_SALE`) as non-editable.
5. **Regression coverage**: added tests for nil/empty selector input and state-only selection behavior when `appStoreState` is missing.

## Validation
- `go test ./internal/cli/migrate -run "TestSelectBestAppInfoID" -count=1`
- `make test`
- `make build`

## Notes
- `make format` was attempted but `gofumpt` is not installed in this environment; files were formatted with `gofmt`.

### DIFF
--- a/internal/cli/migrate/migrate.go
+++ b/internal/cli/migrate/migrate.go
@@ -13,30 +13,6 @@ import (
 	"github.com/rudrankriyam/App-Store-Connect-CLI/internal/asc"
 )
 
-func pickEditableAppInfoID(appInfos *asc.AppInfosResponse) string {
-	if appInfos == nil || len(appInfos.Data) == 0 {
-		return ""
-	}
-
-	// Prefer the draft app info record. Apps commonly have both a live AppInfo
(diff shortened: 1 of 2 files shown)


rudrankriyam:
@cameronehrlich made some changes; what do you think

cameronehrlich:
> @cameronehrlich made some changes; what do you think

Looks good to me!

PR #462, opened 2026-02-09
Add MileIO to Wall of Apps

juergen-kc:
Adds [MileIO](https://apps.apple.com/app/id6758225631) to the Wall of Apps.

MileIO is a fuel and vehicle expense tracker for iOS — tracks gas, diesel, hybrid, and EV charging with receipt scanning, multi-vehicle support, analytics, and Siri Shortcuts integration. Built and shipped entirely with `asc` for TestFlight distribution and App Store submissions.


rudrankriyam:
Awesome! 🎉

PR #466, opened 2026-02-10
fix: complete age rating declaration support and handle whatsNew on initial releases

cameronehrlich:
## Problem

The `age-rating set` command fails when trying to update age rating declarations because the CLI is missing several fields that Apple's API now requires. The API returns:

```
You must provide a value for the attribute 'gunsOrOtherWeapons'
You must provide a value for the attribute 'userGeneratedContent'
You must provide a value for the attribute 'healthOrWellnessTopics'
```

Additionally, uploading version localizations with a `whatsNew` field on an initial v1.0 release causes a hard failure, since Apple doesn't support What's New text until there's a prior version.

Finally, `OptionalBool` flags like `--available-in-new-territories` fail when using space-separated syntax (`--flag true`) because `IsBoolFlag()` tells Go's flag parser to treat them as standalone booleans.

### 1. Add all missing age rating fields

Cross-referenced Apple's OpenAPI spec to add every missing content descriptor:

| Field | Type | Status |
|-------|------|--------|
| `gunsOrOtherWeapons` | enum | **was missing** |
| `advertising` | bool | **new** |
| `healthOrWellnessTopics` | bool | **new** |
| `lootBox` | bool | **new** |
| `messagingAndChat` | bool | **new** |
| `parentalControls` | bool | **new** |
| `ageAssurance` | bool | **new** |
| `userGeneratedContent` | bool | **new** |
| `ageRatingOverride` | string | **new** (response only) |
| `koreaAgeRatingOverride` | string | **new** (response only) |

### 2. Fix output display

All new fields are now shown in `age-rating get --output table` and `--output markdown`.

### 3. Table-driven parsing

Refactored `buildAgeRatingAttributes` from ~130 lines of repetitive blocks into two declarative tables with shared loops. Adding a new field now requires one table entry instead of changes in three places.

### 4. Graceful whatsNew handling

When uploading version localizations for an initial release, the API rejects the whatsNew field. The CLI now catches this, warns the user, and retries without it.

### 5. Fix OptionalBool space-separated flag parsing

`OptionalBool` implemented `IsBoolFlag()` returning `true`, which tells Go's flag package to treat the flag like a built-in bool: `--flag value` is parsed as `--flag` (implicit true) + positional arg. This broke `--available-in-new-territories true` and any other OptionalBool flag using space-separated syntax.

**Fix:** Remove `IsBoolFlag()` entirely so the parser always expects an explicit value. Both `--flag=true` and `--flag true` work correctly after this change.

### 6. Cleanup

- Removed deprecated `--seventeen-plus` flag (not in Apple's current API spec)
- Removed dead nil-checks for fields that were never set by the CLI

## Files changed

| File | What changed |
|------|-------------|
| `internal/asc/age_rating.go` | Added 10 missing struct fields |
| `internal/asc/output_age_rating.go` | Added all new fields to table/markdown output |
| `internal/asc/output_test.go` | Tests verify new fields appear in output |
| `internal/cli/agerating/age_rating.go` | Table-driven parsing, new CLI flags, removed seventeen-plus |
| `internal/cli/agerating/age_rating_test.go` | Tests for bool/enum parsing, invalid values |
| `internal/cli/shared/localizations.go` | whatsNew retry logic |
| `internal/cli/shared/optional_bool.go` | Removed IsBoolFlag() to fix space-separated parsing |

## Test plan

- [x] `go test ./internal/cli/agerating/` - all pass
- [x] `go test ./internal/asc/` - all pass
- [x] `go test ./internal/cli/shared/` - all pass
- [x] `go test ./internal/cli/apps/` - all pass
- [x] Verified age rating set/get against live app
- [x] Verified whatsNew retry on v1.0 submission
- [x] Verified --available-in-new-territories true (space-separated) works after fix

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -11,6 +11,7 @@ LDFLAGS := -X main.version=$(VERSION) -X main.commit=$(COMMIT) -X main.date=$(DA
 GO := go
 GOMOD := go.mod
 GOBIN := $(shell $(GO) env GOPATH)/bin
+GOLANGCI_LINT_TIMEOUT ?= 5m
 INSTALL_PREFIX ?= /usr/local/bin
 
 # Directories
@@ -74,7 +75,7 @@ test-integration:
 lint:
 	@echo "$(BLUE)Linting code...$(NC)"
 	@if command -v golangci-lint >/dev/null 2>&1; then \
-		golangci-lint run ./...; \
+		golangci-lint run --timeout=$(GOLANGCI_LINT_TIMEOUT) ./...; \
 	else \
 		echo "$(YELLOW)golangci-lint not found; falling back to 'go vet ./...'.$(NC)"; \
 		echo "$(YELLOW)Install with: make tools (or: $(GO) install github.com/golangci/golangci-lint/cmd/golangci-lint@latest)$(NC)"; \


rudrankriyam:
Awesome! 🎉

PR #502, opened 2026-02-11
Add --paginate support to subscription price-points equalizations

humblemuzzu:
## Problem

The `subscriptions price-points equalizations` command only returns the first page (50 items) from Apple's API. Since every subscription has 175 territories, the remaining 124+ equalized price points are silently truncated with no way to retrieve them.

This makes it impossible to programmatically set equalized prices for all territories — you can only reach the first 50.

## Fix

Added `--paginate` flag to the equalizations subcommand, using the same `PaginateAll` pattern already used by the `price-points list` command.

### Changes

**`internal/asc/client_subscription_resources.go`**
- `GetSubscriptionPricePointEqualizations` now accepts `...SubscriptionPricePointsOption` so callers can pass a `nextURL` for pagination (same pattern as `GetSubscriptionPricePoints`)

**`internal/cli/subscriptions/price_points.go`**
- Added `--paginate` flag to the equalizations subcommand
- When set, uses `asc.PaginateAll()` to aggregate all pages before output

### Before / After

```bash
# Before: silently returns only 50 of 174 territories
asc subscriptions price-points equalizations --id "PRICE_POINT_ID"
# → 50 items

# After: returns all 174 territories
asc subscriptions price-points equalizations --id "PRICE_POINT_ID" --paginate
# → 174 items
```

Without `--paginate`, behavior is unchanged (backward compatible).

All existing tests pass.

### DIFF
--- a/internal/asc/subscriptions_resources_http_test.go
+++ b/internal/asc/subscriptions_resources_http_test.go
@@ -815,6 +815,50 @@ func TestGetSubscriptionPricePointEqualizations(t *testing.T) {
 	}
 }
 
+func TestGetSubscriptionPricePointEqualizations_WithLimit(t *testing.T) {
+	response := jsonResponse(http.StatusOK, `{"data":[{"type":"subscriptionPricePoints","id":"eq-1"}]}`)
+	client := newTestClient(t, func(req *http.Request) {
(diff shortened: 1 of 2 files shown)


rudrankriyam:
Awesome!

PR #503, opened 2026-02-11
Add Repetti and TV Show Tracker to Wall of Apps

rursache:
Thanks to this amazing CLI tool, appstore submissions are now a blast! keep up the good work! i submited just two of the apps i've used asc with but only to not spam this repo... i'm using asc on a daily basis and on over 10 apps!


rursache:
@rudrankriyam can you please update the `wall-of-apps.json`? i somehow only saw that after i did the PR 😞

PR #660, opened 2026-02-19
feat(testflight): add --internal, --all-builds flags to beta-groups create

ss5nathan:
## Summary

- The `beta-groups create` command currently only supports `--app` and `--name`, always creating **external** beta groups that require Apple's beta review before testers can install
- Adds `--internal`, `--all-builds`, and `--feedback-enabled` flags to `beta-groups create`, following the same `fs.Visit()` pattern already used by `beta-groups update`
- Changes `CreateBetaGroup()` client method to accept a `BetaGroupAttributes` struct instead of a bare `name` string, allowing the new API fields to be passed through

## Motivation

External beta groups require Apple's beta review before testers can install builds. Internal groups allow immediate installation for App Store Connect team members — a critical workflow for indie developers and small teams doing rapid iteration. The Apple API already supports `isInternalGroup: true` in the POST body, and the Go structs already have this field, but the CLI command didn't expose it.

### `internal/cli/testflight/beta_groups.go`
- Added `--internal` (bool, default false) — creates an internal group
- Added `--all-builds` (bool, default false) — grants access to all builds
- Added `--feedback-enabled` (bool, default true) — enables feedback
- Uses `fs.Visit()` to track explicitly-passed flags (same pattern as `update`)
- Updated help text and examples

### `internal/asc/client.go`
- Changed `CreateBetaGroup(ctx, appID, name string)` to `CreateBetaGroup(ctx, appID string, attrs BetaGroupAttributes)`
- The struct fields (`IsInternalGroup`, `HasAccessToAllBuilds`, `FeedbackEnabled`) are serialized via `omitempty`, so unset fields are excluded from the request body

### `internal/asc/client_http_test.go`
- Updated existing `TestCreateBetaGroup_SendsRequest` to use new signature
- Added `TestCreateBetaGroup_InternalGroup` verifying all new attributes are sent in the POST body and parsed in the response

## Test plan

- [x] `go test ./...` passes (all 82 packages)
- [ ] Manual test: `asc testflight beta-groups create --app APP_ID --name "Test" --internal`
- [ ] Manual test: `asc testflight beta-groups create --app APP_ID --name "Test"` (backwards-compatible, external group)

### DIFF
--- a/internal/asc/client.go
+++ b/internal/asc/client.go
@@ -657,11 +657,11 @@ func (c *Client) GetBetaGroupTesters(ctx context.Context, groupID string, opts .
 }
 
 // CreateBetaGroup creates a beta group for an app.
-func (c *Client) CreateBetaGroup(ctx context.Context, appID, name string) (*BetaGroupResponse, error) {
+func (c *Client) CreateBetaGroup(ctx context.Context, appID string, attrs BetaGroupAttributes) (*BetaGroupResponse, error) {
 	payload := BetaGroupCreateRequest{
 		Data: BetaGroupCreateData{
 			Type:       ResourceTypeBetaGroups,
-			Attributes: BetaGroupAttributes{Name: name},
+			Attributes: attrs,
 			Relationships: &BetaGroupRelationships{
 				App: &Relationship{
 					Data: ResourceData{
--- a/internal/asc/client_http_test.go
+++ b/internal/asc/client_http_test.go
@@ -1622,8 +1622,64 @@ func TestCreateBetaGroup_SendsRequest(t *testing.T) {
 		assertAuthorized(t, req)
 	}, response)
 
-	if _, err := client.CreateBetaGroup(context.Background(), "app-1", "Beta"); err != nil {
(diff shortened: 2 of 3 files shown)


rudrankriyam:
Thanks for the contribution, @ss5nathan! Great minds think alike — I had been working on the same feature and just merged #658 which covers all of this:

- `beta-groups create --internal`
- `beta-groups list --internal / --external` filters
- `update --feedback-enabled`

It also removed `--internal` and `--all-builds` from `update` since ASC rejects those attributes on PATCH requests (they're create-only).

Since the functionality is now on `main`, I'm closing this as already covered. Appreciate you taking the time to put this together — feel free to pick up another issue or open a new one!

rudrankriyam:
`--all-builds` is something I need to check again; if you think it works fine in production for your app and not rejected like internal for patch, please rebase and scope it down to it. Thank you for vibing to the sun ☀️

ss5nathan:
Thanks for adding! --internal is all I needed, if I end up using all builds and it works I'll open PR

PR #784, opened 2026-02-25
feat(shots): macOS app capture + Mac App Store canvas framing

ramonclaudio:
Adds Mac App Store screenshot support.

`--provider macos` grabs the frontmost window of a running macOS app by bundle ID using `screencapture -l <windowID>`. Window ID comes from a Swift one-liner piped to `swift -` via `CGWindowListCopyWindowInfo`. No cgo, no extra binaries.

`--device mac` renders to 2880x1800 (`APP_DESKTOP`) without a device bezel. Screenshot scales to fill the canvas. Optional title/subtitle overlays and background color via `--title`, `--subtitle`, `--bg-color`, `--title-color`, `--subtitle-color`. Using canvas flags on a non-canvas device returns an error.

```bash
# Capture
asc screenshots capture --provider macos --bundle-id com.example.MyApp --name home

# Frame
asc screenshots frame \
  --input ./screenshots/raw/home.png \
  --device mac \
  --title "My App" \
  --subtitle "Your tagline" \
  --bg-color "#f5f5f7" \
  --output-dir ./screenshots/framed
```

Running `--provider macos` requires Screen Recording permission for your terminal (Ghostty, iTerm2, Terminal.app) and Xcode Command Line Tools for `swift`. Both are macOS-only.

Also fixed `ASC_TIMEOUT` being silently ignored for `screenshots capture` and `screenshots frame`. Both now use `ContextWithTimeout`.

Tested on a regular window app and a menubar panel app. Results looked good so figured I'd open it. Ping me if anything breaks.

### DIFF
--- a/internal/cli/cmdtest/shots_capture_test.go
+++ b/internal/cli/cmdtest/shots_capture_test.go
@@ -36,7 +36,7 @@ func TestShotsCapture_RequiredFlagErrors(t *testing.T) {
 			wantErr: "--provider must be",
 		},
 		{
-			name:    "provider must be axe",
+			name:    "simctl is not a valid provider",
 			args:    []string{"screenshots", "capture", "--bundle-id", "com.example.app", "--name", "home", "--provider", "simctl"},
 			wantErr: "--provider must be",
 		},
--- a/internal/cli/cmdtest/shots_frame_test.go
+++ b/internal/cli/cmdtest/shots_frame_test.go
@@ -537,6 +537,34 @@ func TestShotsFrame_CanvasFlagsRejectNonCanvasDevice(t *testing.T) {
 	}
 }
 
+func TestShotsFrame_CanvasFlagsRejectConfigMode(t *testing.T) {
+	root := RootCommand("1.2.3")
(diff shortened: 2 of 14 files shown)


rudrankriyam:
Just tried it out; really cool!

image

rudrankriyam:
I am thinking of adding a frame to it, like how we have for iOS and iPad devices. What do you think?

ramonclaudio:
Yea. I tried to keep the PR as lean and un-opinionated as possible so I went without the frame, but we can deff add a '-frame' flag and allow the user to choose. Koubou already has the frames, they just need to be dropped in.

https://github.com/bitomule/Koubou/tree/main/src/koubou/frames

rudrankriyam:
I think it would be nice to have the screenshots inside the frame and the text outside, like how it is with the iphone devices. I would love if you can add it! 

I also moved the presets to Koubou as a PR so the commands act as a wrapper around it

rudrankriyam:
I am merging this as it is; can have a follow-up PR for the frames. Awesome work!

PR #788, opened 2026-02-26
feat: Support selecting price points by customer price in schedule creation

### LINKED ISSUE #649: Pricing: support tier-based selection for price points
## Problem
Several commands require an opaque `price-point` ID (or `PRICE_POINT_ID` inside `--prices` entries). In practice, people think in terms of a *tier* (sorted list of price points for a base territory like USA) or a *customer price* (e.g. 4.99).

Today, to pick a price point you have to:
- list price points, export JSON, and manually map/sort
- or maintain a manual tier table that must be regenerated when changing the base price/territory

Examples of affected commands:
- `asc pricing schedule create --app APP_ID --price-point PRICE_POINT_ID --base-territory USA --start-date 2024-03-01`
- `asc subscriptions prices add --id SUB_ID --price-point PRICE_POINT_ID [--territory USA]`
- `asc iap price-schedules create --iap-id IAP_ID --base-territory USA --prices PRICE_POINT_ID:2024-03-01`

## Desired behavior
Allow specifying a *tier* (and/or a *customer price*) instead of a raw price point ID.

Important note: Apple no longer exposes official numbered tiers in the ASC API. This would be a CLI-defined convenience layer built on top of price points.

### 1) Compute tiers dynamically (API-derived)
- Fetch price points for a given territory (e.g. USA).
- Sort by `customerPrice` (numeric) ascending.
- Assign tier numbers starting at 1.
- Expose `{tier, id, customerPrice, proceeds, ...}` in output.

### 2) Add tier/price resolution flags (mutually exclusive)
For commands that currently accept a price point ID, add one (or both) of:
- `--tier <int>`: resolve to the corresponding price point ID for `--territory` (or inferred base territory where applicable)
- `--price <decimal>`: resolve by exact customer price match for `--territory`

Keep `--price-point` as the escape hatch.

### 3) Provide a way to list tiers
Add a listing view that prints tier + price point mapping for a territory. Options:
- new command like `asc pricing tiers --app APP_ID --territory USA`
- or extend existing `asc pricing price-points` with a `--with-tiers`/`--tiers` output mode

### 4) Caching + refresh
To avoid repeated API calls and make this ergonomic:
- cache tier mapping per resource kind + territory (and possibly app/subscription/iap context) in a local cache file
- add `--refresh` (or TTL) to re-fetch
- include `asOf` metadata in JSON so tier tables are reproducible

## Acceptance criteria
- Can run `asc pricing schedule create` using `--tier` or `--price` instead of `--price-point`.
- Can run `asc subscriptions prices add` using `--tier` or `--price` instead of `--price-point`.
- Can list tiers for a territory in a fast, human-friendly output (`--output table` / `--output markdown`) and JSON.
- Validation: `--tier`, `--price`, `--price-point` are mutually exclusive; invalid tier/price produces exit code 2 with a clear error.
- Tests: add `internal/cli/cmdtest` coverage with `httptest` fixtures.

## Notes / considerations
- Tier numbering is inherently derived; if Apple adds/removes price points, tier numbers may shift.
- Reference: Apple’s “App Store Pricing Update” PDF (price steps/ranges + conventions, not a tier list): https://www.apple.com/newsroom/pdfs/App-Store-Pricing-Update.pdf
- Default territory should probably be explicit for tier resolution to avoid accidentally triggering very large requests (esp. subscriptions without `--territory`).

WZBbiao:
This PR adds a `--price` flag to the `pricing schedule create` command, allowing users to select a price point by its customer price value (e.g., "0.99") instead of the opaque Price Point ID.\n\nWhen `--price` is provided:\n1. The CLI fetches all price points for the specified base territory.\n2. It finds the price point with a matching `customerPrice`.\n3. It uses that price point's ID to create the schedule.\n\nThis resolves #649 by providing a more user-friendly way to set prices, aligning with the "tier-based" selection request (as tiers map to prices).\n\nUsage:\n`asc pricing schedule create --app <ID> --base-territory USA --price "0.99" --start-date "2026-03-01"`

### DIFF
--- a/internal/cli/cmdtest/pricing_schedule_create_price_test.go
+++ b/internal/cli/cmdtest/pricing_schedule_create_price_test.go
@@ -1,0 +1,166 @@
+package cmdtest
+
+import (
+	"context"
+	"encoding/json"
+	"errors"
+	"flag"
+	"io"
+	"net/http"
+	"strings"
+	"testing"
+)
+
+func TestPricingScheduleCreatePriceValidationError(t *testing.T) {
+	root := RootCommand("1.2.3")
+	root.FlagSet.SetOutput(io.Discard)
+
+	stdout, stderr := captureOutput(t, func() {
+		if err := root.Parse([]string{
+			"pricing", "schedule", "create",
+			"--app", "app-1",
+			"--price", "abc",
+			"--base-territory", "USA",
+			"--start-date", "2026-03-01",
+		}); err != nil {
+			t.Fatalf("parse error: %v", err)
+		}
+		err := root.Run(context.Background())
+		if !errors.Is(err, flag.ErrHelp) {
+			t.Fatalf("expected ErrHelp, got %v", err)
+		}
+	})
+
+	if stdout != "" {
+		t.Fatalf("expected empty stdout, got %q", stdout)
+	}
+	if !strings.Contains(stderr, "Error: --price must be a number") {
+		t.Fatalf("expected invalid number error, got %q", stderr)
+	}
+}
+
+func TestPricingScheduleCreatePriceValidationFiniteError(t *testing.T) {
+	root := RootCommand("1.2.3")
+	root.FlagSet.SetOutput(io.Discard)
+
+	stdout, stderr := captureOutput(t, func() {
+		if err := root.Parse([]string{
+			"pricing", "schedule", "create",
+			"--app", "app-1",
+			"--price", "NaN",
+			"--base-territory", "USA",
+			"--start-date", "2026-03-01",
+		}); err != nil {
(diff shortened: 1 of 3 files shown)


rudrankriyam:
Follow-up pushed to harden --price behavior:

- validate --price as numeric/finite and return usage-style errors
- switch price-point resolution from string equality to numeric matching logic
- add coverage for invalid price inputs and numeric-equivalent resolution (0.990 matching 0.99)

Given #649 scope, this is best framed as partial progress rather than full resolution.

rudrankriyam:
I can merge it as it is, but it does not fully resolve #649 as I want to *hardcode* the tier pricing for ALL countries, which is a **big** task

PR #804, opened 2026-02-28
Add Weather mini to Wall of Apps

kailuo:
- Added Weather mini - Trip Forecasts to Wall of Apps in docs/wall-of-apps.json


rudrankriyam:
🚀

PR #862, opened 2026-03-05
feat(install): add cached skills update notifier

### LINKED ISSUE #797: Feature: Skills Update Notification
## Problem

Users who have installed skills via `asc install-skills` have no way to know when the skills repository has been updated with new capabilities, bug fixes, or improved instructions. This leads to users missing out on improved automation workflows.

## Proposed Solution

Add a subtle update check that runs on CLI invocation (after skills are detected as installed) and shows a non-blocking notification when new skills versions are available.

### 1. Version Tracking
Store the installed skills version/commit in `~/.asc/config.json` alongside existing config:
```json
{
  "skills_version": "abc123",
  "skills_checked_at": "2026-02-27T10:00:00Z"
}
```

### 2. Update Check Mechanism
- Use GitHub API to fetch the latest commit SHA from `rudrankriyam/app-store-connect-cli-skills`
- Compare with stored version
- Only check once per day (cache the check timestamp)

### 3. UI/UX
- Non-blocking banner printed to stderr (doesn't interfere with JSON output)
- Subtle formatting, e.g.:
```
⚡ asc-skills v1.2.0 available. Run 'asc install-skills' to update.
```
- Respects `ASC_DEBUG` flag (silent when debug enabled)

### 4. Implementation Location
- Add skills tracking fields to `internal/config/config.go`
- Create new `internal/cli/install/skills_check.go` for update logic
- Hook into `cmd/root.go` Exec function (runs before any command)

### 5. Configuration
- Add `ASC_SKILLS_AUTO_CHECK` env var to disable (default: enabled)
- Cache results in config to avoid network calls on every invocation

## Trade-offs Considered

- **Could use npm registry instead of GitHub API** - but GitHub provides better commit-level tracking for the skills repo
- **Could auto-update skills silently** - but explicit update via `asc install-skills` is safer and gives users control
- **Could block on update check** - but async/subtle is better UX to avoid slowing down CLI usage

## Expected Invocation Examples

After installing skills, the first run after an update would show:
```
$ asc builds list
⚡ asc-skills v1.2.0 available. Run 'asc install-skills' to update.
...
```

## Edge Cases & Failure Modes

- Network failure during check: silently skip (don't block CLI usage)
- No skills installed: don't show any message
- User has disabled auto-check: respect `ASC_SKILLS_AUTO_CHECK=false`
- Skills repo is private or unavailable: graceful degradation

---

This is a community-maintained, unofficial skill pack and is not affiliated with Apple.

rudrankriyam:
## Summary
- Add a future-proof skills update notification path that delegates update detection to the upstream `skills check` command instead of hardcoded repository commit polling.
- Run the check at most once per day (cached via `skills_checked_at` in config), emit a non-blocking stderr hint only when updates appear available, and never fail user commands on check errors.
- Add `ASC_SKILLS_AUTO_CHECK` support (`true/false`, `1/0`, `yes/no`, `y/n`, `on/off`) and wire the check into command execution for real subcommands while skipping root-only invocation and `install-skills` itself.

## Test plan
- [x] `go test ./internal/cli/install ./cmd`
- [x] `make format`
- [x] `make generate-command-docs`
- [x] `make lint`

Closes #797

### DIFF
--- a/internal/cli/install/skills_check.go
+++ b/internal/cli/install/skills_check.go
@@ -9,6 +9,7 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
+	"sort"
 	"strings"
 	"time"
 
@@ -31,6 +32,7 @@ var (
 	runSkillsCheckCommand          = defaultRunSkillsCheckCommand
 	progressEnabledForCheck        = shared.ProgressEnabled
 	lookupSkillsCheckCLI           = exec.LookPath
+	errSkillsCheckUnavailable      = errors.New("skills check command unavailable")
 )
 
 // MaybeCheckForSkillUpdates checks for skill updates once per interval and prints
@@ -67,7 +69,9 @@ func MaybeCheckForSkillUpdates(ctx context.Context) {
 	if runErr != nil {
 		// Avoid suppressing future checks when the command never actually ran due
 		// to cancellation or timeout in the parent context.
-		if !errors.Is(runErr, context.Canceled) && !errors.Is(runErr, context.DeadlineExceeded) {
+		if !errors.Is(runErr, context.Canceled) &&
+			!errors.Is(runErr, context.DeadlineExceeded) &&
+			!errors.Is(runErr, errSkillsCheckUnavailable) {
 			_ = persistSkillsCheckedAtForCheck(now.Format(skillsCheckedAtLayout))
 		}
 		return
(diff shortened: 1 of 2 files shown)


rudrankriyam:
Addressed the review feedback in 2482b231: (1) auto skill checks are now explicit opt-in by default, (2) singular 'no update available' no longer false-positives, and (3) redundant npx lookup error branching removed. Ran format/lint/tests before pushing.

rudrankriyam:
Updated intent: ASC_SKILLS_AUTO_CHECK is intentionally default-enabled. To address the security concern while keeping that default, this push runs the check command outside the project directory and keeps --no-install, so untrusted repo-local node_modules binaries are not auto-resolved.

PR #870, opened 2026-03-05
Fix audit findings around output, auth, config, and release safety

rudrankriyam:
## Summary
- validate shared output flags before command execution so invalid `--output` and `--pretty` combinations fail with a usage error before auth or API side effects
- align auth resolution with documented precedence by preferring stored credentials, stopping env fallback on stored-source errors, and making invalid strict/bypass env settings conservative
- harden config persistence with atomic symlink-safe writes and gate release publishing on format, docs, lint, and tests

## Why this approach
- wrapping shared output validation at the command layer fixes the bug once for the command tree instead of relying on each mutating command to remember to prevalidate before calling `PrintOutput`
- changing auth resolution was preferable to updating docs because the current behavior could silently switch accounts when stored credentials were broken
- atomic config writes reuse the repo's secure-open primitives instead of layering more ad hoc file checks into `SaveAt`

## Alternatives considered
- keep late output validation inside render helpers and patch mutating commands one by one
  - rejected because it leaves the same class of regression open for future commands
- preserve current auth fallback behavior and document it more clearly
  - rejected because it keeps the wrong-account risk and conflicts with the intended precedence contract

## Invocation notes
- `asc devices register ... --output yaml` now fails with usage exit code `2` before auth or API calls
- `asc devices update ... --output table --pretty` now fails with usage exit code `2` before any mutation path runs
- broken stored credentials now return the stored-source error instead of silently falling back to env credentials

## Verification
- `make format`
- `make check-command-docs`
- `make lint`

## Exit-code scenarios validated
- invalid `--output` returns usage before auth
- invalid `--pretty`/format combinations return usage before auth or mutation
- stored-credential errors stop env fallback
- config writes reject symlink targets safely

### DIFF
--- a/internal/cli/shared/shared.go
+++ b/internal/cli/shared/shared.go
@@ -608,7 +608,7 @@ func strictAuthEnabled() bool {
 		return false
 	default:
 		warnInvalidStrictAuthValueOnce(value)
-		return true
+		return false
 	}
 }
 
@@ -627,7 +627,7 @@ func warnInvalidStrictAuthValueOnce(value string) {
 
 	fmt.Fprintf(
 		os.Stderr,
-		"Warning: invalid %s value %q (expected true/false, 1/0, yes/no, y/n, or on/off); strict auth enabled conservatively\n",
+		"Warning: invalid %s value %q (expected true/false, 1/0, yes/no, y/n, or on/off); strict auth disabled\n",
 		strictAuthEnvVar,
 		value,
 	)
--- a/internal/cli/shared/shared_test.go
+++ b/internal/cli/shared/shared_test.go
@@ -898,7 +898,7 @@ func TestStrictAuthEnabled_EnvFalseyValues(t *testing.T) {
 	}
 }
 
-func TestStrictAuthEnabled_InvalidValueWarnsAndEnables(t *testing.T) {
+func TestStrictAuthEnabled_InvalidValueWarnsAndDisables(t *testing.T) {
 	previousStrict := strictAuth
 	strictAuth = false
 	t.Cleanup(func() {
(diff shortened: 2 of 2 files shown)


rudrankriyam:
Follow-up update on top of the initial audit fix set:

- parent-scoped output validation now checks the full command path, so flags like `asc reviews --output yaml ratings ...` fail before the leaf command runs
- invalid `ASC_BYPASS_KEYCHAIN` and `ASC_STRICT_AUTH` values now resolve conservatively; both warn and do not enable bypass/strict mode
- config writes now sync the relevant directories around temp-file creation and final rename, with regression coverage for the sync path

Additional verification:
- `ASC_BYPASS_KEYCHAIN=1 go test ./cmd -run 'TestRun_Invalid(Output|Pretty|ParentOutput|ParentPretty)ReturnsUsageBefore(LeafExec|Auth)?'`
- `ASC_BYPASS_KEYCHAIN=1 go test ./internal/auth -run 'TestShouldBypassKeychainEnvSemantics|TestShouldBypassKeychain_InvalidValueWarnsAndDisables'`
- `ASC_BYPASS_KEYCHAIN=1 go test ./internal/cli/shared -run 'TestStrictAuthEnabled_(EnvTruthyValues|EnvFalseyValues|InvalidValueWarnsAndDisables)'`
- `ASC_BYPASS_KEYCHAIN=1 go test ./internal/config -run 'TestSaveAt(OverwritesExistingFile|RejectsSymlinkPath|SyncsDirectoriesDuringWrite|PropagatesDirectorySyncErrors)'`
- `go build -o /tmp/asc-audit . && /tmp/asc-audit reviews --output yaml ratings --app 1479784361` (exits with usage code 2 and prints `unsupported format: yaml`)

PR #721, opened 2026-02-21
feat: add Swift helper integration for enhanced iOS workflow support

dl-alexandre:
## Summary

This PR integrates the swift-helpers toolchain into the GitHub Actions pipelines and updates the contributor checklist to run tests with ASC_BYPASS_KEYCHAIN=1.  ￼

## What changed
	-	Updates the testing checkbox to use ASC_BYPASS_KEYCHAIN=1 make test.  ￼
	-	CI workflows
	-	main-branch.yml and pr-checks.yml now build Swift helpers and add their build output to PATH.
	-	PR checks now run a targeted test set focused on Swift helper integration and related validation paths (instead of only go test -short ./...).  ￼
	-	Release workflow
	-	Adds a guard step to detect whether Apple Developer signing secrets are configured.
	-	Imports signing certs only when secrets are present.
	-	Builds Swift helpers (release config), includes them in the release directory, codesigns them alongside the main binaries (when configured), and includes them in checksums.
	-	Packages Swift helpers into a versioned tarball artifact.  ￼
	-	New workflow: swift-helpers.yml
	-	Adds dedicated CI for Swift helpers: build (debug/release), artifact upload, Swift tests, Go integration tests exercising helpers, benchmarks, and release-only signing/notarization stages.  ￼

## Why
	-	Ensures Swift helper binaries are consistently built and validated in CI, and shipped correctly with releases.
	-	Reduces CI flakiness and local dev friction by standardizing ASC_BYPASS_KEYCHAIN=1 for relevant test runs.  ￼

## How to test
	-	CI:
	-	Verify pr-checks.yml builds Swift helpers and runs the narrowed Go test set.
	-	Verify swift-helpers.yml runs build/test/integration jobs on PRs touching Swift helper paths.
	-	Local:
	-	ASC_BYPASS_KEYCHAIN=1 make test
	-	cd swift-helpers && swift test
	-	swift build -c release && go test ./internal/swifthelpers/...  ￼

## Notes
	-	Release signing is now conditional: if Apple Developer secrets are not configured, the workflow proceeds without codesigning and prints a warning.  ￼

PR #904, opened 2026-03-07
feat: add Swift helper integration for enhanced iOS workflow support

rudrankriyam:
Replacement for #721 with a clean replay from current `main`.

## Summary
- Replay the meaningful Swift-helper commit series from #721 on top of current `main`, preserving original author names and commit messages via `git cherry-pick -x`.
- Keep the reviewed Swift-helper scope only: packaging support, native-framework helpers, benchmarks, performance work, and the follow-up removals of the Swift keychain and daemon scope.
- Drop obsolete merge commits and stale CI/lint branch-maintenance churn so the PR diff reflects the real feature work instead of the old fork `main` history.

## Test plan
- [x] `make format`
- [x] `make lint`

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -60,6 +60,21 @@ Canonical test rule: all test runs must use `ASC_BYPASS_KEYCHAIN=1` to avoid hos
 - CI must enforce formatting + lint + tests on both PR and `main` workflows.
 - Remove unused shared wrappers/helpers when commands are refactored.
 
+## PR Audit Workflow
+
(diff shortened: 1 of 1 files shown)


rudrankriyam:
Closing this for now @dl-alexandre 

I think this is better as a plugin or optional companion package than something we merge into the core CLI. It adds a lot of surface area to `asc` itself: Swift toolchain/setup, extra binaries, more CI/release complexity, and a separate install/discovery story.

We can explore this in #962  and `asc plugins` style path later. Thanks again for the work here!

PR #949, opened 2026-03-11
refactor: make app-info list the canonical discovery path

### LINKED ISSUE #948: Consolidate `app-infos` into the canonical `asc app-info` command family
## Outcome
Consolidate `asc app-infos ...` into the canonical `asc app-info ...` family so app metadata discovery has one obvious home.

## Why This Matters
This is a small but high-value cleanup because it removes a singular/plural split in a real metadata workflow. The point is better discovery and fewer mental branches, not renaming for its own sake.

## Current State
Relevant files:

- `internal/cli/apps/app_info.go`
- `internal/cli/apps/app_infos.go`
- `internal/cli/apps/app_info_relationships.go`
- `internal/cli/apps/app_info_territory_age_ratings.go`
- `internal/cli/registry/registry.go`

Today:

- `asc app-info ...` owns the main metadata surface
- `asc app-infos list` exists as a separate plural root mostly to help users discover an app info ID

That creates one domain with two names.

## Improvement Direction
- add `asc app-info list` as the canonical listing and discovery path
- keep `asc app-infos list` only as a deprecated compatibility alias during transition
- update help, docs, and error guidance to point to `app-info`
- preserve the discoverability benefit of the current list workflow

Possible canonical shape:

- `asc app-info list --app "APP_ID"`
- `asc app-info get ...`
- `asc app-info set ...`
- `asc app-info relationships ...`
- `asc app-info territory-age-ratings ...`

## Guardrails
- do not lose the listing/discovery affordance that the plural root currently provides
- do not keep both names equally prominent after the canonical path exists
- keep output and validation semantics stable across the migration
- make the metadata surface easier to explain, not just more internally consistent

## Definition Of Done
- `asc app-info list` exists and is the documented discovery path
- `asc app-infos list` remains only as a compatibility alias for at least one deprecation cycle
- help/examples/error guidance consistently point to `app-info`
- the registry no longer treats `app-infos` as a first-class domain
- cmdtests cover canonical help/discovery plus alias behavior
- any command-doc snapshots are regenerated if help text changes

## Notes
This is a good candidate for a narrow, reviewable change because it improves metadata taxonomy without requiring a broad behavior rewrite.

rudrankriyam:
## Summary
- add `asc app-info list` as the canonical app info discovery path under the existing `app-info` family
- keep `asc app-infos list` working as a hidden deprecated compatibility alias with the same runtime behavior and a canonical warning
- remove `app-infos` from primary root/docs discovery and add cmdtests for canonical help, hidden alias behavior, and output parity

## Test plan
- [x] `make format`
- [x] `make lint`

Closes #948.

### DIFF
--- a/cmd/root_usage.go
+++ b/cmd/root_usage.go
@@ -32,7 +32,7 @@ var rootUsageGroups = []rootCommandGroup{
 	{
 		title: "APP MANAGEMENT COMMANDS",
 		commands: []string{
-			"apps", "app-setup", "app-tags", "app-info", "app-infos", "versions",
+			"apps", "app-setup", "app-tags", "app-info", "versions",
 			"localizations", "screenshots", "video-previews", "background-assets", "product-pages",
 			"routing-coverage", "pricing", "pre-orders", "categories", "age-rating",
 			"accessibility", "encryption", "eula", "agreements", "app-clips",
--- a/docs/COMMANDS.md
+++ b/docs/COMMANDS.md
@@ -59,7 +59,6 @@ asc <subcommand> [flags]
 - `app-setup` - Post-create app setup automation.
 - `app-tags` - Manage app tags for App Store visibility.
 - `app-info` - Manage App Store version metadata.
-- `app-infos` - List app info records for an app.
 - `versions` - Manage App Store versions.
 - `localizations` - Manage App Store localization metadata.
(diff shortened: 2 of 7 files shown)


rudrankriyam:
Superseded by a broader  redesign that removes the old  /  split instead of only canonicalizing it.

PR #953, opened 2026-03-11
refactor: finish builds test-notes canonical migration

### LINKED ISSUE #945: Complete migration from `beta-build-localizations` to `asc builds test-notes`
## Summary
The repository now has a better canonical home for TestFlight What to Test notes at `asc builds test-notes ...`, but the old top-level `beta-build-localizations` command group is still registered for compatibility. We should finish that migration so discovery points to the builds-centric surface first.

## Current state in code
Relevant files:

- `internal/cli/builds/build_test_notes.go`
- `internal/cli/builds/builds_commands.go`
- `internal/cli/betabuildlocalizations/beta_build_localizations.go`
- `internal/cli/betabuildlocalizations/beta_build_localizations_build.go`
- `internal/cli/registry/registry.go`

The new canonical surface already exists:

- `asc builds test-notes list|get|create|update|delete`

At the same time, the legacy root still exists and even documents itself as deprecated:

- `asc beta-build-localizations ...`

The deprecation note in `BetaBuildLocalizationsCommand()` explicitly says users should prefer `asc builds test-notes ...`.

## Problem
The migration is only partially complete:

- users can still discover the legacy root from the registry
- the command tree still has two homes for the same workflow
- the older root includes behavior (`--global`) that does not map cleanly onto the new builds-centric taxonomy
- long-term maintenance has to keep both surfaces in sync

## Proposed direction
Make `asc builds test-notes ...` the only canonical discovery path.

Possible migration approach:

1. Keep build-scoped note management under `asc builds test-notes ...`
2. Decide the final home for the current global-list behavior from `beta-build-localizations list --global`
   - either keep it as legacy-only during deprecation
   - or expose a canonical equivalent elsewhere if it is still valuable
3. Convert `beta-build-localizations` into a hidden/deprecated compatibility alias layer
4. Remove it from primary docs/help examples once the alias coverage is in place

## Open design question
`beta-build-localizations list --global` is the one piece that does not naturally fit a build-scoped `test-notes` group. The implementation should explicitly choose one of these paths:

- keep `--global` legacy-only until removal
- add a canonical global listing surface under `builds`
- decide that the global listing is not worth preserving as a first-class path

The issue is done only when that choice is documented and tested.

## Acceptance criteria
- `asc builds test-notes ...` is the documented canonical path
- `asc beta-build-localizations ...` is clearly deprecated and hidden from normal discovery
- the future of `--global` is explicitly defined
- docs/examples stop teaching the legacy root
- cmdtests cover canonical behavior and the remaining alias behavior

## Suggested test coverage
- canonical help tests for `builds test-notes`
- legacy deprecation/runtime tests for `beta-build-localizations`
- tests covering the chosen treatment of `--global`
- upload + `--test-notes` integration behavior remains intact

## Why now
This migration is already conceptually underway in the codebase. Finishing it will reduce surface-area duplication and make builds the obvious home for build-scoped TestFlight note workflows.

rudrankriyam:
## Summary
- make `asc builds test-notes ...` the canonical What to Test surface and rename the canonical read path from `get` to `view`
- keep `asc builds test-notes get` and `asc beta-build-localizations ...` as deprecated compatibility aliases while hiding the legacy root from normal discovery and docs
- preserve legacy-only selectors like `--global`, `--app --latest`, and `build get` on the deprecated root during the migration window

## Test plan
- [x] `make format`
- [x] `make generate-command-docs`
- [x] `make lint`

Closes #945.

### DIFF
--- a/docs/COMMANDS.md
+++ b/docs/COMMANDS.md
@@ -88,7 +88,6 @@ asc <subcommand> [flags]
 - `pre-release-versions` - Manage TestFlight pre-release versions.
 - `build-localizations` - Manage build release notes localizations.
 - `beta-app-localizations` - Manage TestFlight beta app localizations.
-- `beta-build-localizations` - Manage TestFlight beta build localizations.
 - `sandbox` - Manage sandbox testers in App Store Connect.
 
 ### Review and Release
--- a/internal/cli/betabuildlocalizations/beta_build_localizations.go
+++ b/internal/cli/betabuildlocalizations/beta_build_localizations.go
@@ -20,28 +20,53 @@ func BetaBuildLocalizationsCommand() *ffcli.Command {
 	return &ffcli.Command{
 		Name:       "beta-build-localizations",
 		ShortUsage: "asc beta-build-localizations <subcommand> [flags]",
-		ShortHelp:  "Manage TestFlight beta build localizations.",
-		LongHelp: `Manage TestFlight beta build localizations ("What to Test" notes).
+		ShortHelp:  "DEPRECATED: use `asc builds test-notes ...`.",
+		LongHelp: `Deprecated compatibility layer for TestFlight What to Test notes.
 
-Deprecated: prefer "asc builds test-notes ..." for the same What to Test functionality.
-This command group remains available for compatibility.
+Canonical build-scoped workflows now live under ` + "`asc builds test-notes ...`" + `.
+
+Legacy-only behaviors still remain here during the transition:
+- ` + "`beta-build-localizations list --global`" + `
+- ` + "`beta-build-localizations get --app ... --latest`" + `
+- ` + "`beta-build-localizations create --app ... --latest`" + `
+- ` + "`beta-build-localizations build get`" + `
 
 Examples:
-  asc beta-build-localizations list --build "BUILD_ID"
-  asc beta-build-localizations list --global
-  asc beta-build-localizations list --global --paginate
-  asc beta-build-localizations create --build "BUILD_ID" --locale "en-US" --whats-new "Test instructions"`,
+  asc builds test-notes list --build "BUILD_ID"
+  asc builds test-notes view --id "LOCALIZATION_ID"
+  asc builds test-notes create --build "BUILD_ID" --locale "en-US" --whats-new "Test instructions"`,
 		FlagSet:   fs,
-		UsageFunc: shared.DefaultUsageFunc,
+		UsageFunc: shared.DeprecatedUsageFunc,
 		Subcommands: []*ffcli.Command{
-			BetaBuildLocalizationsListCommand(),
-			BetaBuildLocalizationsGetCommand(),
+			deprecatedBetaBuildLocalizationsLeafCommand(
+				BetaBuildLocalizationsListCommand(),
+				"asc builds test-notes list",
+				"Warning: `asc beta-build-localizations list` is deprecated. Use `asc builds test-notes list` for build-scoped workflows. `--global` remains legacy-only during transition.",
+			),
+			deprecatedBetaBuildLocalizationsLeafCommand(
+				BetaBuildLocalizationsGetCommand(),
+				"asc builds test-notes view",
+				"Warning: `asc beta-build-localizations get` is deprecated. Use `asc builds test-notes view` for ID-based lookups. `--latest` remains legacy-only during transition.",
+			),
 			BetaBuildLocalizationsBuildCommand(),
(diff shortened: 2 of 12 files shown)


Thaina:
Are there any reason `--app --latest` feature was removed from the new `test-notes` command?

And is there any alternative?

rudrankriyam:
@Thaina I checked the history here.

`--app --latest` was not removed from `asc builds test-notes` itself; that shortcut was never part of the canonical `builds test-notes` surface.

Timeline:
- `asc builds test-notes` was introduced on 2026-01-25 and from the start it used explicit `--build` / `--id` selectors.
- `--app --latest --state` was added later on 2026-03-05, but only to the legacy `asc beta-build-localizations get/create` flow, mainly to help post-upload workflows where the build ID is not known yet.
- In this migration PR, the intent was to keep `asc builds test-notes ...` as the canonical, explicit, build-scoped surface, while preserving legacy-only selectors like `--app --latest` on the deprecated `beta-build-localizations` command during the transition.

Alternatives today:
- Keep using the legacy shortcut:
  - `asc beta-build-localizations get --app "APP_ID" --latest --state "PROCESSING,COMPLETE" --locale "en-US"`
  - `asc beta-build-localizations create --app "APP_ID" --latest --state "PROCESSING,COMPLETE" --locale "en-US" --whats-new "..."`
- Or use the canonical two-step flow:
  - `asc builds latest --app "APP_ID" --processing-state "PROCESSING,VALID"`
  - then `asc builds test-notes ... --build "BUILD_ID"`

Small nuance: for `asc builds latest`, the equivalent of old `COMPLETE` is `VALID`.

Thaina:
@rudrankriyam I am sorry I did not know at the time I request that feature, that the `beta-build-localizations` was already deprecated and we should use `builds test-notes`

However I still thought that even with the new command, the ability to infer latest build without explicit build number is still very useful, it should not be deprecated along with `beta-build-localizations`

In fact, there is already many subcommand utilize this `ResolveBuild` logic

https://github.com/rudrankriyam/App-Store-Connect-CLI/blob/cdf62e553298bf75162233455c4cfe0fe6d4a3d8/internal/cli/builds/resolve_build.go#L21-L25

I wish this all of subcommand, such as `test-notes` and `info`, that require build ID should use this same logic so that I could specify `--latest` to all of them in the same manner

And actually we should deprecate `asc builds latest --app "123"` and redirected to `asc builds info --latest --app "123"`

Is it possible?

rudrankriyam:
Yes, good idea. Will be there in the new build tonight~

rudrankriyam:
I think there are more inconsistencies here; thank you for bringing it to my attention. Will do a better refactor across all the commands so it is easier to work with

Thaina:
Also I see there is `wait` that use `--newest` instead of `--latest`

Is it the same logic and should it reuse same flag?

>  there are more inconsistencies here; thank you for bringing it to my attention. Will do a better refactor across all the commands so it is easier to work with

I see, very appreciate your amazing work

rudrankriyam:
> Also I see there is wait that use --newest instead of --latest. Is it the same logic and should it reuse same flag?

Yep. I am cleaning up the commands and the underlying duplicated code so the news ones will look like: 

```bash
asc builds info --app "123" --latest
asc builds wait --app "123" --latest
asc builds dsyms --app "123" --build-number "42"
asc builds test-notes create --app "123" --latest --locale "en-US"
```

rudrankriyam:
I decided to support one selector model across build-targeting `asc builds` commands:

```bash
asc builds info --app "123" --latest --output json
asc builds expire --app "123" --latest --confirm
asc builds update --app "123" --build-number "42" --version "1.2.3" --platform IOS --uses-non-exempt-encryption=false
```

I originally split this as:

- read/query flows: allow inferred selection
- mutating/destructive flows: require `--build-id`

I have changed that direction. That split made the CLI less coherent and forced users and automation to remember special cases inside one command family.

The final selector contract is:

- `--build-id`: direct identity
- `--app --latest`: ranked selection
- `--app --build-number`: unique lookup

Important details:

- `--latest` is the only ranked selector
- `--build-number` must resolve to exactly one build after filters, otherwise the CLI errors
- there is no implicit `IOS` default for unique lookup; pass `--platform` when needed
- mutating/destructive commands still use their normal `--confirm` flags where appropriate, but they do not have a stricter selector than read commands

So yes: mutating build commands now support the same selector too!

Thaina:
that's reasonable. Thank you for your effort

rudrankriyam:
Sorry for the AI generated comment; been working all day on this and tired

PR #955, opened 2026-03-11
feat: validate release configuration (type and scheduled date)

Abdo-codes:
## Summary

- Adds release configuration validation to `asc validate`
- `release.scheduled_date_past` (warning): catches SCHEDULED releases with an earliest release date in the past — the app would release immediately after approval instead of waiting, which may not be the developer's intent
- `release.type_manual` (info): reminds that MANUAL release requires the developer to explicitly trigger release after approval via `asc versions release`
- Adds `ReleaseType` and `EarliestReleaseDate` fields to `AppStoreVersionAttributes` so the version GET response surfaces these fields (previously only available in create/update requests)

## Test plan

- [x] Unit tests for all release check scenarios (past date, future date, manual, after_approval, empty, case-insensitive)
- [x] Integration test: `TestValidateWarnsScheduledReleaseDateInPast` — version with SCHEDULED + past date triggers warning
- [x] Integration test: `TestValidateShowsManualReleaseInfo` — version with MANUAL triggers info check
- [x] Wired through `Validate()` — `TestValidateIncludesReleaseChecks`
- [x] No regressions: full `TestValidate*` suite passes

### DIFF
--- a/internal/asc/client_versions.go
+++ b/internal/asc/client_versions.go
@@ -9,11 +9,13 @@ import (
 
 // AppStoreVersionAttributes describes app store version metadata.
 type AppStoreVersionAttributes struct {
-	Platform        Platform `json:"platform,omitempty"`
-	VersionString   string   `json:"versionString,omitempty"`
-	AppStoreState   string   `json:"appStoreState,omitempty"`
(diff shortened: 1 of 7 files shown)


rudrankriyam:
🔥

PR #999, opened 2026-03-12
fix: exclude territory resources from included array during availability create

### LINKED ISSUE #998: pricing availability set: fails to CREATE availability for apps without existing record
## Bug

**asc version:** 0.40.1

**Command:**
```
asc pricing availability set --app 6759846474 --territory "USA" --available true --available-in-new-territories true
```

**Error:**
```
Error: pricing availability set: app availability not found for app "6759846474" and failed to create: The provided entity id is invalid: The provided included entity id 'USA' has invalid format. For inline creation, the id must be a local id with the format '${local-id}'.
```

**Expected:** Should create an availability record for apps that don't have one yet.

**Context:** This works fine for apps that already have an availability record (e.g. updating territories). The issue is only when creating availability for the first time. The Apple v2 `appAvailabilities` create endpoint likely expects territory IDs in a local-id format (`${territory-USA}`) rather than plain codes (`USA`).

chadnewbry:
## Problem

`asc pricing availability set` (and `asc app-setup availability set`) fails when creating availability for apps that don't have an existing availability record.

**Error:**
```
The provided included entity id 'AFG' has invalid format.
For inline creation, the id must be a local id with the format '${local-id}'.
```

## Root Cause

The `CreateAppAvailabilityV2` function adds bare territory resources (`{type: "territories", id: "AFG"}`) to the `included` array. Apple's v2 API rejects non-local IDs in `included` during inline creation.

Territories are existing resources — they should only be referenced via the `territory` relationship inside the inline-created `territoryAvailability` resources, not included separately.

## Fix

Remove the territory resources from the `included` array. The territory availabilities already reference territories via their nested `relationships.territory.data` — that's sufficient for the API.

## Testing

Tested against a real app (`6759846474`) that had no prior availability record:
- Before fix: fails with `invalid format` error
- After fix: successfully creates availability for all 175 territories

Fixes #998

### DIFF
--- a/internal/asc/client_pricing.go
+++ b/internal/asc/client_pricing.go
@@ -525,7 +525,8 @@ func (c *Client) GetAppAvailabilityV2TerritoryAvailabilitiesRelationships(ctx co
 	return &response, nil
 }
 
-// CreateAppAvailabilityV2 creates or updates app availability.
+// CreateAppAvailabilityV2 calls POST /v2/appAvailabilities.
+// Apple documents this endpoint as app pre-order creation, not generic app-availability bootstrap.
 func (c *Client) CreateAppAvailabilityV2(ctx context.Context, appID string, attrs AppAvailabilityV2CreateAttributes) (*AppAvailabilityV2Response, error) {
 	appID = strings.TrimSpace(appID)
 	if appID == "" {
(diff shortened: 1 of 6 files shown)


rudrankriyam:
@chadnewbry ok I really found some bigger problems that I have to fix. Thank you so much for creating the issue!

rudrankriyam:
I pushed a follow-up commit onto this branch to keep the public availability commands update-only.

What I verified is that the original change here does not fix the root issue for apps that have no existing availability record.

The core problem is that the public create path we were using for `pricing availability set` is not the same contract as the App Store Connect UI bootstrap flow:

- the public `POST /v2/appAvailabilities` path rejects the UI-style `availableTerritories` relationship
- the public path also does not successfully bootstrap the missing record through the `territoryAvailabilities` payloads we were trying to reshape here
- because of that, removing `territories` from `included` cannot make the missing-record case work reliably

This branch now does the safer thing on the public path:

- keep `pricing availability set` / `app-setup availability set` focused on existing availability records
- stop attempting a public auto-create when the app has no availability record yet
- clarify the command guidance and add regression coverage for the update-only behavior

Checks run on the branch after the change:

- `make lint`
- `make check-command-docs`

PR #1043, opened 2026-03-13
feat: recover submit create when version is in another submission

Abdo-codes:
## Summary

- When `submit create` fails because the version is already in another review submission, it now recovers by extracting the existing submission ID from the error and submitting that instead
- Cleans up the empty submission it created before recovering
- Addresses a common pain point where stale/orphaned submissions from prior failed attempts block new submissions

## Problem

`asc submit create` fails with an opaque error when the version was already added to another review submission (e.g. from a prior failed attempt or manual App Store Connect usage):

```
appStoreVersions with id X was already added to another reviewSubmission with id Y
```

The user is left to manually find and submit the existing submission.

## Solution

1. Parse the `ENTITY_ERROR` associated error to extract the existing submission UUID
2. Reuse the existing submission by submitting it directly
3. Cancel the empty new submission to avoid orphans
4. Log the recovery to stderr so the user knows what happened

## Test plan

- [x] Unit tests for `extractExistingSubmissionID` (5 cases: happy path, non-API error, no matching detail, no associated errors, wrapped error)
- [x] Integration test mocking the full recovery flow (API returns 409, recovery extracts ID, cancels empty submission, submits existing one)
- [x] All 3262 existing tests pass with no regressions

### DIFF
--- a/internal/cli/cmdtest/submit_create_test.go
+++ b/internal/cli/cmdtest/submit_create_test.go
@@ -937,6 +937,139 @@ func TestSubmitCreateRecoversFromAlreadyAddedError(t *testing.T) {
 	}
 }
 
+func TestSubmitCreateRetriesWhenConflictPointsToRecentlyCanceledStaleSubmission(t *testing.T) {
+	setupSubmitCreateAuth(t)
+
+	originalTransport := http.DefaultTransport
+	t.Cleanup(func() {
+		http.DefaultTransport = originalTransport
+	})
+
+	requests := make([]string, 0, 12)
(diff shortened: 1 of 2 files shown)


rudrankriyam:
Bro insane

Abdo-codes:
Addressed Codex feedback: updated the regex to use `(?i)` for case-insensitivity and `\S+` instead of a strict UUID pattern, since the API spec defines `ReviewSubmission.id` as a generic string. Added test cases for uppercase UUIDs and non-UUID identifiers. Also rebased on main to resolve the merge conflict.

PR #1044, opened 2026-03-13
feat: auto-select editable app info when multiple exist

Abdo-codes:
## Summary

- `ResolveAppInfoID` now auto-selects the most editable app info instead of erroring when an app has multiple (e.g. READY_FOR_SALE + PREPARE_FOR_SUBMISSION)
- Uses the existing `SelectBestAppInfoID` helper which prefers PREPARE_FOR_SUBMISSION, then any non-live state, then falls back to the first
- Logs the auto-selected ID to stderr for transparency

## Problem

When an app has two app infos (common during updates), every command that resolves app info fails:

```
multiple app infos found for app "123"; run `asc apps info list` ...
```

This forces users to manually look up and pass `--app-info` on every invocation of `categories set`, `metadata push`, `localizations update`, etc.

## Solution

Reuse the existing `SelectBestAppInfoID()` (already used by `validate readiness` and `migrate import`) inside `ResolveAppInfoID()`. The selection priority is:

1. `PREPARE_FOR_SUBMISSION` (the editable one)
2. Any non-live state (IN_REVIEW, WAITING_FOR_REVIEW, etc.)
3. First app info (fallback when all are live)

## Test plan

- [x] 6 new unit tests covering: explicit override, single app info, auto-select editable, auto-select non-live, fallback to first when all live, no app infos error
- [x] Full shared package test suite passes (285 tests)

### DIFF
--- a/internal/cli/cmdtest/apps_info_surface_test.go
+++ b/internal/cli/cmdtest/apps_info_surface_test.go
@@ -204,3 +204,64 @@ func TestDeprecatedAppInfosListAliasWarnsAndMatchesAppsInfoListOutput(t *testing
 		t.Fatalf("expected canonical and alias output to match, canonical=%q alias=%q", canonicalStdout, aliasStdout)
 	}
 }
+
+func TestAppsInfoViewIncludeFailsWhenAppInfoIsAmbiguous(t *testing.T) {
+	setupAuth(t)
+	t.Setenv("ASC_BYPASS_KEYCHAIN", "1")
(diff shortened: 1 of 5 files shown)


rudrankriyam:
🤔

Abdo-codes:
it's just better ux?

rudrankriyam:
I am thinking why I do not find such stuff 🤣

Abdo-codes:
God's plan 🤣

PR #1045, opened 2026-03-13
feat: add `asc auth token` command to print signed JWT

Abdo-codes:
## Summary
- Adds `asc auth token` subcommand that prints a signed App Store Connect JWT to stdout
- Raw token output (no newline) for direct piping into `curl` or other tools
- Supports `--output json` for structured output with token, keyId, and profile name
- Supports `--name` flag to select a specific credential profile
- Follows existing credential resolution chain (default profile > single stored > error on ambiguity)

## Motivation
When the CLI doesn't support a particular API endpoint or you need a quick escape hatch for debugging, there's currently no way to get a valid JWT without writing custom code. This command fills that gap:

```bash
# Direct API call
curl -H "Authorization: Bearer $(asc auth token)" \
  https://api.appstoreconnect.apple.com/v1/apps

# JSON output with metadata
asc auth token --output json

# Specific profile
asc auth token --name "CI-Key"
```

## Test plan
- [x] No credentials → clear error message
- [x] Profile not found → clear error message
- [x] Raw token output is valid JWT (3 dot-separated parts)
- [x] JSON output includes token, keyId, profile
- [x] Named profile selection works
- [x] Multiple credentials without `--name` → error with guidance
- [x] Unexpected positional args rejected
- [x] `resolveTokenCredential` unit tests for default and named resolution
- [x] All 57 existing auth tests still pass

### DIFF
--- a/cmd/run_additional_test.go
+++ b/cmd/run_additional_test.go
@@ -667,6 +667,54 @@ func TestRun_AuthTokenAmbiguousProfilesReturnAuthExit(t *testing.T) {
 	}
 }
 
+func TestRun_AuthTokenRejectsPermissiveKeyFile(t *testing.T) {
+	resetReportFlags(t)
+
+	tempDir := t.TempDir()
+	configPath := filepath.Join(tempDir, "config.json")
+	keyPath := filepath.Join(tempDir, "AuthKey.p8")
+	writeRunTestECDSAPEM(t, keyPath)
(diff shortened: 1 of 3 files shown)


rudrankriyam:
Hmm need to think about security and explicitly use --confirm

Abdo-codes:
hmm, asc developer mood?

rudrankriyam:
For such, I prefer --confirm so if it leaks or something, it is on the developer

PR #1074, opened 2026-03-15
fix: suppress 409 Conflict warnings on stale submission cancel

Abdo-codes:
## Summary

- When `submit create` cancels stale READY_FOR_REVIEW submissions before creating a new one, a race condition can cause the cancel to hit a 409 Conflict (submission already transitioned to a non-cancellable state between list and cancel)
- Previously this printed alarming `Warning: failed to cancel stale submission` messages to stderr even though the situation is harmless
- Now silently skips 409 Conflict errors while still warning on genuine failures (502, 500, etc.)

## Test plan

- [x] New test `TestSubmitCreateSilentlySkipsConflictOnStaleSubmissionCancel` — verifies 409 on cancel produces no warning
- [x] Existing `TestSubmitCreateWarnsWhenStaleSubmissionCancelFails` — still warns on 502
- [x] Full test suite passes (`ASC_BYPASS_KEYCHAIN=1 make test`)

### DIFF
--- a/internal/cli/cmdtest/submit_create_test.go
+++ b/internal/cli/cmdtest/submit_create_test.go
@@ -516,12 +516,12 @@ func TestSubmitCreateSilentlySkipsConflictOnStaleSubmissionCancel(t *testing.T) 
 		}
 	})
 
-	// 409 Conflict on stale cancel should be silently suppressed — no warning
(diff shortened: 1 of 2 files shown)


rudrankriyam:
Wow

PR #1087, opened 2026-03-15
feat(age-rating): add --all-none flag for safe defaults

Abdo-codes:
## Summary
Adds `--all-none` flag to `asc age-rating set` that sets all 22 rating attributes to safe defaults (NONE/false) in one command.

## Problem
Setting age rating for a simple app with no objectionable content requires typing 22 individual flags. This is the most common case — most apps just need everything set to NONE.

## Solution
```bash
# Before: 22 flags
asc age-rating set --app APP_ID --gambling false --violence-realistic NONE --sexual-content-nudity NONE ...

# After: 1 flag
asc age-rating set --app APP_ID --all-none
```

Individual flags override `--all-none` when both are provided.

## Test plan
- [x] `go build ./...` passes
- [x] `go test ./internal/cli/agerating/...` passes

### DIFF
--- a/cmd/age_rating_run_test.go
+++ b/cmd/age_rating_run_test.go
@@ -23,3 +23,22 @@ func TestRun_AgeRatingSetInvalidAllNoneReturnsUsage(t *testing.T) {
 		t.Fatalf("expected invalid boolean value error, got %q", stderr)
 	}
 }
+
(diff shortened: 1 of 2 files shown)


Abdo-codes:
Review comments addressed: The xcode.go and version.go comments are from a stale review of the previous branch state. This PR now contains only 1 commit modifying `internal/cli/agerating/age_rating.go`. Force-pushed to clean branch based on upstream/main.

PR #1088, opened 2026-03-15
fix(xcode): handle tvOS direct-upload exports without IPA

Abdo-codes:
## Summary
Fixes false error on `asc xcode export` when exporting tvOS apps or using `destination: upload` in ExportOptions.

## Problem
When exporting a tvOS archive with `destination: upload`, xcodebuild uploads directly to App Store Connect without producing a local `.ipa` file. The CLI then errors with "xcodebuild export did not produce an .ipa file" even though the upload succeeded.

## Solution
- Reads ExportOptions.plist to detect `destination: upload` mode
- When in upload mode, skips IPA file search after successful xcodebuild
- Returns bundle info from the archive instead
- CLI shows "(direct upload -- no local artifact)" for IPA path

## Test plan
- [x] `go build ./...` passes
- [x] `go test ./internal/xcode/...` passes
- [x] `go test ./internal/cli/xcode/...` passes

### DIFF
--- a/docs/wall-of-apps.json
+++ b/docs/wall-of-apps.json
@@ -516,6 +516,11 @@
     "icon": "https://is1-ssl.mzstatic.com/image/thumb/Purple211/v4/87/cf/44/87cf44ee-79ee-1a91-3edc-d195ad86582b/AppIcon-0-0-1x_U007emarketing-0-8-0-85-220.png/512x512bb.jpg"
   },
   {
+    "app": "Space - Life Planner",
+    "link": "https://apps.apple.com/us/app/space-life-planner/id6502294348?uo=4",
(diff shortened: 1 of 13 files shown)


Abdo-codes:
Review feedback addressed in force-push:
1. **prepareIPAPath data loss** — Upload mode is now detected *before* `prepareIPAPath`, so existing IPAs are never deleted when `destination=upload`.
2. **Silent metadata error** — `readArchiveBundleInfo` errors are now propagated instead of silently discarded.

Abdo-codes:
Addressed new feedback:
- **Parent dir creation**: Now extracted before the uploadMode check — `os.MkdirTemp` always has a valid parent dir.

PR #1090, opened 2026-03-16
feat(apps): add content-rights view/edit commands

Abdo-codes:
## Summary
Adds `asc apps content-rights view/edit` commands to manage the `contentRightsDeclaration` attribute required for App Store submission.

## Problem
When submitting an app, Apple requires `contentRightsDeclaration` to be set. There was no focused CLI command for this, so users had to go to the App Store Connect web UI or use a broader app update flow.

## Usage
```bash
# Check current value
asc apps content-rights view --app APP_ID

# Set it (most apps)
asc apps content-rights edit --app APP_ID --uses-third-party-content=false

# If app uses third-party content
asc apps content-rights edit --app APP_ID --uses-third-party-content=true
```

## Implementation
- Uses existing `UpdateApp` API method and `ContentRightsDeclaration` type
- Follows the canonical `view/edit` command shape used elsewhere in the CLI
- Renders the declaration directly so default table output shows the value being read or updated
- No new API methods needed

## Test plan
- [x] `go build ./...` passes
- [x] `go test ./internal/cli/apps/...` passes

### DIFF
--- a/internal/cli/apps/apps.go
+++ b/internal/cli/apps/apps.go
@@ -53,7 +53,7 @@ Examples:
   asc apps update --id "APP_ID" --bundle-id "com.example.app"
   asc apps update --id "APP_ID" --primary-locale "en-US"
   asc apps subscription-grace-period get --app "APP_ID"
-  asc apps content-rights set --app "APP_ID" --uses-third-party-content=false
+  asc apps content-rights edit --app "APP_ID" --uses-third-party-content=false
   asc apps --limit 10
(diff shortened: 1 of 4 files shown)


Abdo-codes:
Addressed all feedback:
- **Explicit value required**: Changed from bool to string flag — empty value errors. Accepts `true/false/yes/no/uses/does-not-use` and raw API values
- **Boolean false syntax**: No longer relevant — it's a string flag now

PR #1126, opened 2026-03-18
feat: add `asc builds count` command

thatswiftguy:
## Summary

- Adds `asc builds count --app APP_ID` to report the total number of builds for an app without fetching the full list
- Uses `meta.paging.total` from the API response with `limit=1` — a single fast request regardless of how many builds exist
- Adds `ParsePagingTotalOK(meta) (int, bool)` to correctly distinguish "zero builds" from "API did not return a total"

## How It Works

Apple's `GET /v1/builds` response includes `meta.paging.total` in its pagination metadata. By setting `limit=1` internally, we make the smallest possible request and read the total count from metadata rather than fetching all builds.

## Filters Supported

Same filters as `asc builds list`:

| Flag | Description |
|---|---|
| `--version` | Filter by marketing version (CFBundleShortVersionString) |
| `--build-number` | Filter by build number (CFBundleVersion) |
| `--platform` | Filter by `IOS`, `MAC_OS`, `TV_OS`, `VISION_OS` |
| `--processing-state` | Filter by `VALID`, `PROCESSING`, `FAILED`, `INVALID`, or `all` |

## Example Output

```bash
$ asc builds count --app 6753645132
┌────────────┬──────────────┐
│   App ID   │ Total Builds │
├────────────┼──────────────┤
│ 6753645132 │ 2            │
└────────────┴──────────────┘

$ asc builds count --app 6753645132 --output json
{"appID":"6753645132","total":2}

### DIFF
--- a/internal/cli/builds/builds_commands.go
+++ b/internal/cli/builds/builds_commands.go
@@ -619,8 +619,10 @@ func findPreReleaseVersionIDsForBuildsList(
 		for _, preReleaseVersion := range page.Data {
 			// ASC's version filter can return dotted-version near-matches like
 			// 1.1 and 1.1.0 together, so enforce exact matching client-side
-			// before converting pre-release versions into build filters.
-			if strings.TrimSpace(preReleaseVersion.Attributes.Version) != version {
+			// when the response includes attributes.version. If ASC omits the
+			// attribute entirely, trust the server-side filter instead.
+			versionAttr := strings.TrimSpace(preReleaseVersion.Attributes.Version)
+			if versionAttr != "" && versionAttr != version {
 				continue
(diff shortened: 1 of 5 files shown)


rudrankriyam:
What is the use-case for it?

thatswiftguy:
> What is the use-case for it?

It can be used to quickly get the build count, specially when a build is under process on test flight.

rudrankriyam:
I ended up finding a bug in `builds list` too while testing; fixing

thatswiftguy:
> I ended up finding a bug in `builds list` too while testing; fixing

Oh, let me know, I can try fixing it.

Issue #1180, opened 2026-03-21
web auth: surface session expiry and investigate non-relogin renewal


mithileshchellappan:
## Summary

`asc web auth login` currently creates a short-lived Apple web session and requires an explicit relogin once the cached cookies expire. We should expose the exact cached expiry/TTL and investigate whether we can extend or renew the session without asking the user to log in again.

## Observed behavior

- Current cached sample:
  - issued at `2026-03-16 17:04:33 UTC` (`2026-03-16 22:34:33 IST`)
  - expires at `2026-03-17 01:04:33 UTC` (`2026-03-17 06:34:33 IST`)
  - original TTL: `8h`
- The cached cookie jar stores session cookies without explicit `Expires`, but the embedded payloads expose timing:
  - `dqsid.iat`
  - `itctx.ex`
- Once expired, `GET /olympus/v1/session` returns `401 Unauthenticated` and `TryResumeSession` surfaces `ErrCachedSessionExpired`.

## Current implementation notes

- `internal/web/auth.go` sends `rememberMe: false` and `signin/complete?isRememberMeEnabled=false`.
- `internal/web/session_cache.go` validates cached sessions by replaying cookies against `/olympus/v1/session`.
- If Apple rotates cookies during that request, the current code will already persist the new cookies, but there is no proactive renewal path and no evidence yet that Apple actually rotates them for this flow.

## Why this matters

- `asc web` commands are already experimental and fragile.
- An ~8 hour session lifetime is short enough to break real multi-step workflows in the same day.
- Today users get a generic expired-session outcome instead of a precise expiry/TTL and a clear explanation of what can or cannot be renewed.

## Investigation tracks

1. Expose exact expiry metadata in `asc web auth status`.
   - Parse `dqsid.iat` and `itctx.ex` without printing token values.
   - Emit issued-at, expires-at, and remaining TTL.

2. Test whether the login flow can request longer-lived cookies.
   - Compare the current `rememberMe: false` / `isRememberMeEnabled=false` flow with `true`.
   - Verify whether Apple returns persistent cookies or a meaningfully longer TTL for App Store Connect web sessions.

3. Test whether a still-valid session can be renewed without full relogin.
   - Probe `/olympus/v1/session` and any adjacent low-risk authenticated endpoints for `Set-Cookie` rotation near expiry.
   - If rotation happens, add a proactive refresh path before `itctx.ex`.

4. Decide what to do if true renewal is not possible.
[...]

PR #1181, opened 2026-03-21
web: add lazy auto-reauth for expired sessions

rudrankriyam:
## Summary
- add one-shot lazy auto-reauth for expired `asc web` sessions when a password is already available
- reuse the cached cookie jar for relogin instead of throwing away Apple trust cookies and starting from a blank jar
- persist the refreshed session after successful relogin and cover the new flow with web CLI + session cache tests

## Why this approach
Live validation on March 21, 2026 pointed away from token refresh and toward cached-jar relogin:
- an expired ASC session returned `GET /olympus/v1/session -> 401`
- the same trusted browser/profile still held long-lived Apple trust state (`DES...`)
- re-entering only email + password recreated the ASC session with no 2FA prompt
- the relogin flow was `signin/init -> 200`, `signin/complete -> 200`, `GET /olympus/v1/session -> 200`
- no `/verify/...` or `/2sv/trust` calls appeared in that relogin run

This PR mirrors that behavior in the CLI by preserving the cached jar during relogin instead of claiming a non-existent refresh-token path.

## Alternatives considered
- background refresh / keepalive: rejected because we have no credible refresh-token path and the short-lived ASC session cookies do not justify timer churn
- always restart from a blank jar: rejected because it throws away the trust cookies that appear to be the whole reason password-only relogin can skip repeated 2FA
- add keychain-backed Apple Account password storage in the same PR: deferred to keep this change focused; the current implementation only auto-reauths when a password is already available

## Behavior
Examples:
```bash
ASC_WEB_PASSWORD='...' asc web review list --app 123456789 --apple-id user@example.com
ASC_WEB_PASSWORD='...' asc web auth login --apple-id user@example.com
```

On cache hit:
- existing cached session is reused as before

On expired cache:
- the CLI tries one auto-reauth using the cached jar and available password source
- on success the original command continues with source `auto-reauth`
- if no password source is available, behavior falls back to the existing explicit login path
- if Apple asks for 2FA again, the existing 2FA handling still applies

## Tested failure modes / edge cases
- expired cached session still prints the old "Session expired." notice before interactive prompt when no silent password source exists
- expired user-scoped cached session auto-reauths with `ASC_WEB_PASSWORD` and reuses the cached client
- expired last cached session can auto-reauth using stored email metadata from the cache
- cached session load helpers hydrate the jar without validating against the live session endpoint so trust cookies survive relogin

## Commands run
- `go test ./internal/cli/web -run 'TestResolveSession(AutoReauthsExpiredCachedSessionUsingEnvPassword|AutoReauthsExpiredLastCachedSessionUsingStoredEmail|PrintsExpiredNoticeBeforePrompt)'`
- `go test ./internal/web -run 'TestLoad(CachedSession|LastCachedSession)HydratesJarWithoutValidation'`
- `go test ./internal/cli/web`
- `go test ./internal/web`
- `make format`
- `make check-command-docs`
- `make lint`

## Notes
- This PR intentionally does not add TTL/expiry metadata to `asc web auth status`; that can land separately if we want to keep the auth-recovery change easy to review.
- Refs #1180.

### DIFF
--- a/internal/cli/web/web_auth.go
+++ b/internal/cli/web/web_auth.go
@@ -220,14 +220,8 @@ func printExpiredSessionNotice(writer io.Writer) {
 }
 
 func loginWithOptionalTwoFactor(ctx context.Context, appleID, password, twoFactorCode string) (*webcore.AuthSession, error) {
-	return loginWithOptionalTwoFactorUsing(ctx, "Signing in to Apple web session", appleID, password, twoFactorCode, func(ctx context.Context, creds webcore.LoginCredentials) (*webcore.AuthSession, error) {
-		return webLoginFn(ctx, creds)
-	})
-}
-
-func loginWithOptionalTwoFactorUsing(ctx context.Context, label, appleID, password, twoFactorCode string, loginFn func(context.Context, webcore.LoginCredentials) (*webcore.AuthSession, error)) (*webcore.AuthSession, error) {
-	session, err := withWebSpinnerValue(label, func() (*webcore.AuthSession, error) {
-		return loginFn(ctx, webcore.LoginCredentials{
+	session, err := withWebSpinnerValue("Signing in to Apple web session", func() (*webcore.AuthSession, error) {
+		return webLoginFn(ctx, webcore.LoginCredentials{
 			Username: appleID,
 			Password: password,
 		})


mithileshchellappan:
@codex review

PR #1183, opened 2026-03-21
web: replace broken web 2fa input with command-based retrieval

rudrankriyam:
## Summary
- replace the broken experimental web `--two-factor-code` flow with `--two-factor-code-command`
- add `ASC_WEB_2FA_CODE_COMMAND` as the env-based fallback for fresh web-session logins
- keep shared web-session test hooks compatible with older resolver stubs while moving user-facing commands to the new path

## Why this approach
The current web 2FA path expects manual terminal input, but that flow is not working in practice for this experimental web auth surface. A command-based handoff lets `asc` fetch the code from AppleScript or another local helper at the exact moment Apple requires 2FA, which matches the intended macOS automation use case.

## Alternatives considered
- Keep `--two-factor-code` and add the command path alongside it. I did not take this path because the existing manual flow is broken and would keep a misleading interface alive.
- Add a file-based 2FA source instead of a command. That would work, but a command is a better fit for AppleScript/UI automation because it can wait for the dialog and print the code only when ready.

## Expected usage
```bash
asc web auth login --apple-id user@example.com --two-factor-code-command "osascript /path/to/get-apple-2fa-code.scpt"
ASC_WEB_PASSWORD="..." ASC_WEB_2FA_CODE_COMMAND="osascript /path/to/get-apple-2fa-code.scpt" asc web apps create --name "My App" --bundle-id com.example.app --sku MYAPP123 --apple-id user@example.com
```

## Edge cases covered
- removed legacy `--two-factor-code` from the user-facing web auth surface
- fail with a clear usage hint when 2FA is required but no command/env provider is configured
- trim command stdout and reject empty output
- surface command stderr when code retrieval fails
- preserve progress labels and cached-session behavior

## Verification
- `ASC_BYPASS_KEYCHAIN=1 go test ./internal/cli/web -run 'TestLoginWithOptionalTwoFactorUsesCommandWhenConfigured|TestResolveSessionUsesTwoFactorCodeCommandEnvWhen2FARequired|TestLoginWithOptionalTwoFactorReturnsPromptError|TestReadTwoFactorCodeFromCommand|TestLoginWithOptionalTwoFactorUsesProvidedCodeWhenPresent|TestLoginWithOptionalTwoFactorUsesProgressLabels|TestResolveSessionUsesProgressLabels'`
- `ASC_BYPASS_KEYCHAIN=1 go test ./internal/cli/cmdtest -run 'TestWebAuthLoginDoesNotExposePlaintextPasswordFlag'`
- `ASC_BYPASS_KEYCHAIN=1 go test ./cmd -run 'TestWebAuthLoginLegacyTwoFactorFlagExitCode'`
- `ASC_BYPASS_KEYCHAIN=1 go test ./internal/cli/web ./internal/cli/cmdtest ./cmd`
- `make format`
- `make check-command-docs`
- `PATH="$(go env GOPATH)/bin:$PATH" make lint`

### DIFF
--- a/internal/cli/web/web_auth_script_test.go
+++ b/internal/cli/web/web_auth_script_test.go
@@ -66,10 +66,27 @@ func TestAppleTwoFactorScriptRestrictsFallbackToRecognizedTrustPrompts(t *testin
 	if !strings.Contains(script, "property trustDialogTextHints :") {
 		t.Fatalf("expected trust dialog text hints in script")
 	}
+	if strings.Contains(script, "\"trust\"") || strings.Contains(script, "\"trusted\"") {
+		t.Fatalf("expected trust dialog hints to avoid generic trust/trusted substrings")
+	}
 	if !strings.Contains(script, "return my windowContainsAnyTextHint(theWindow, trustDialogTextHints)") {
 		t.Fatalf("expected trust dialog detection helper in script")
 	}
 	if !strings.Contains(script, "if not (my looksLikeTrustDialog(theWindow)) then") || !strings.Contains(script, "\t\treturn false") {
 		t.Fatalf("expected script to refuse unknown FollowUpUI dialogs before fallback clicking")
+	}
+}
+
+func TestAppleTwoFactorScriptDoesNotTreatTrustedDevicesPromptAsTrustDialog(t *testing.T) {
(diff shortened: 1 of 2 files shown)


rudrankriyam:
@codex review

PR #1220, opened 2026-03-25
feat: add --free flag to set app price to Free ($0)

Abdo-codes:
## Summary
Adds `--free` flag to the pricing set command, allowing apps to be set to Free ($0) via CLI.

## Problem
Setting an app's price to Free via `asc app-setup pricing set --price "0"` failed with:
```
no price point found matching price 0 in this territory
```
This happened because `ResolveTiers` intentionally excludes $0 price points (they're not pricing tiers). The `--tier 0` flag also didn't work because tier 0 was interpreted as "not provided" (default int value).

During a real shipping session, we needed to switch an app from paid to freemium (free download + IAP subscriptions) and had to fall back to the App Store Connect web UI.

## Solution
New `--free` boolean flag that resolves the $0 price point directly from the API:
```bash
asc app-setup pricing set --app "APP_ID" --free
asc app-setup pricing set --app "APP_ID" --free --base-territory "USA"
```

## Changes
- **`pricing_command.go`**: Add `--free` flag, default `--base-territory` to USA when free, resolve via `ResolveFreeAppPricePoint`
- **`tier_resolver.go`**: Add `ResolveFreeAppPricePoint` — fetches price points and finds the $0 entry. Update `ValidatePriceSelectionFlags` to accept `free` as a 4th mutually exclusive option (variadic to preserve backward compatibility)
- **`tier_resolver_test.go`**: Update error message assertions

## Test plan
- [x] `go build ./...` passes
- [x] `go test ./internal/cli/shared/...` passes
- [ ] Manual: `asc app-setup pricing set --app "APP_ID" --free` sets price to $0
- [ ] Manual: `--free` with `--price` shows mutual exclusivity error
- [ ] Manual: `--free` without `--base-territory` defaults to USA


rudrankriyam:
Lol Codex has such blunt comment, good work @Abdo-codes ! 🚀

Issue #1272, opened 2026-03-29
Feature: add resolved output mode for pricing list commands


mithileshchellappan:
## Summary

add resolved output mode for pricing list commands

## Reproduction

```bash
asc subscriptions pricing prices list --subscription-id 6759909879 --paginate --output json
```

## Expected behavior

List commands should have an easy way to return resolved fields for the common inspection case. For pricing lists, that means territory, amount, currency, dates, and whether a value is manual or equalized, either by default or via a flag like --expand/--resolved.

## Actual behavior

```
The pricing list output is useful as an API-shaped response, but for operational use it still required extra interpretation to answer the simple question of what prices are currently live by territory.
```

## Environment

- **asc version:** 0.47.0 (commit: 5086317, date: 2026-03-27T09:09:02Z)
- **OS:** darwin/arm64
- **Filed via:** `asc snitch`

rudrankriyam:
Closing this out via merged PR #1285: https://github.com/rudrankriyam/App-Store-Connect-CLI/pull/1285

That PR implemented the requested resolved pricing view by adding an opt-in `--resolved` mode for pricing list commands across subscriptions, app price schedules, and IAP price schedules. It closes the gap from this issue by returning the current effective price per territory while keeping the existing raw list output unchanged when `--resolved` is not used.

It also added the pagination/query support needed for app schedule price resolution, preserved required include/fields query params across resolved pagination, and rejects `--resolved --next` because a mid-stream cursor cannot produce a complete current-by-territory result.

This shipped to `main` in merge commit `a49068a3` on 2026-03-30.

PR #1285, opened 2026-03-30
add resolved current pricing views for pricing lists

mithileshchellappan:
## Summary
- add an opt-in `--resolved` view for pricing list commands across subscriptions, IAP schedule prices, and app schedule prices
- keep existing raw list output unchanged when `--resolved` is not set
- add app schedule pagination/query support needed to resolve current effective prices per territory

## Why
Issue #1272 asks for a current-by-territory pricing view from list commands. The existing commands exposed raw history or one-territory summaries, but not a resolved current pricing view across territories.

## Review Fixes
- preserve required `include` and `fields[...]` query params across internal pagination in resolved mode
- reject `--resolved --next` because a mid-stream cursor cannot produce a complete current-by-territory answer
- add regression coverage for both cases

## Impact
- `--resolved` now returns one current effective row per territory for the supported pricing list surfaces
- raw JSON/table/markdown behavior remains unchanged for existing scripts
- app schedule manual/automatic list commands now support `--limit`, `--next`, and `--paginate`

## Validation
- `make format`
- `make check-command-docs`
- `make lint` is blocked in this environment because the installed `golangci-lint` binary was built with Go 1.24 while the repo targets Go 1.26.0

### DIFF
--- a/internal/cli/iap/resolved_schedule_prices.go
+++ b/internal/cli/iap/resolved_schedule_prices.go
@@ -40,7 +40,6 @@ func fetchResolvedIAPSchedulePrices(
 	case "manual":
 		if err := consumeResolvedIAPSchedulePrices(
 			ctx,
-			client,
 			scheduleID,
 			nextURL,
 			limit,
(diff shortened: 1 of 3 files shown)


mithileshchellappan:
@codex review

PR #1294, opened 2026-03-30
[codex] fix testflight testers group filter

### LINKED ISSUE #1292: bug: `testflight testers list --group` fails with "Only one relationship filter"
## Bug

\`asc testflight testers list --app APP_ID --group GROUP_ID\` fails:

```
Error: testers list: failed to fetch: A parameter has an invalid value: Only one relationship filter can be applied.
```

## Root cause

The CLI sends both \`filter[apps]=APP_ID\` and \`filter[betaGroups]=GROUP_ID\` as query parameters. The App Store Connect API only allows one relationship filter at a time on the \`GET /v1/betaTesters\` endpoint.

## Expected

When \`--group\` is provided, the CLI should:
1. Use **only** \`filter[betaGroups]=GROUP_ID\` (drop the app filter)
2. Or use the relationship endpoint directly: \`GET /v1/betaGroups/{GROUP_ID}/betaTesters\`

## Reproduction

\`\`\`bash
# Get a group ID first
asc testflight groups list --app 6747745091 --output json | jq '.data[0].id'
# → "3ce80ef2-925e-4a36-b6d1-5475868714eb"

# This fails:
asc testflight testers list --app 6747745091 --group "3ce80ef2-925e-4a36-b6d1-5475868714eb"
# Error: Only one relationship filter can be applied

# This also fails (group name instead of ID):
asc testflight testers list --app 6747745091 --group "Alpha"

# This works (no group filter):
asc testflight testers list --app 6747745091 --limit 1
# → returns all testers for the app
\`\`\`

## Impact

- Cannot get tester count per beta group
- ASC Studio TestFlight view cannot show per-group tester counts
- The \`--group\` flag is effectively broken when \`--app\` is also set (which is always, since it defaults from env)

rudrankriyam:
## Summary
- stop `testflight testers list` from sending `filter[apps]` when `--group` is present
- reject `--group` with `--build-id` up front instead of letting the API reject conflicting relationship filters
- add CLI and lower-level query coverage for the new precedence and validation rules

## Root Cause
`asc testflight testers list --group ...` resolved the beta group ID and then still called `/v1/betaTesters` with both `filter[apps]` and `filter[betaGroups]`. The App Store Connect API only accepts one relationship filter on that endpoint, so the request failed with `Only one relationship filter can be applied.`

## Why This Approach
I kept the command on `/v1/betaTesters` and changed filter precedence instead of switching to `/v1/betaGroups/{id}/betaTesters`.

That preserves support for `--email`, which the relationship endpoint does not expose, while still producing a valid request for group-scoped listing.

## Alternatives Considered
- Use `GET /v1/betaGroups/{id}/betaTesters`: simpler for pure group listing, but it drops `--email` support and changes the command's filtering behavior.
- Keep the old request shape and rely on API errors: avoids code changes, but leaves the documented `--group` workflow broken.

## Validation
- `go test ./internal/cli/cmdtest -run 'TestTestFlightBetaTestersListWithGroupUsesOnlyGroupFilter|TestTestFlightBetaTestersListRejectsGroupAndBuildID'`
- `go test ./internal/asc -run 'TestBuildBetaTestersQueryPrefersGroupFilterOverAppFilter|TestGetBetaTesters_WithBuildFilter'`
- `go test ./internal/cli/cmdtest -run 'Test(TestFlightBetaTesters|BetaTestersListAcceptsBuildFilter)'`
- `go test ./internal/asc -run 'Test(BuildBetaTestersQuery|GetBetaTesters_)'`
- `go build -o /tmp/asc .`
- built binary exit-code check: `/tmp/asc testflight testers list --app app-1 --group group-1 --build-id build-1` exits with code `2`
- live smoke test: `ASC_BYPASS_KEYCHAIN=1 /tmp/asc testflight testers list --app 6747745091 --group 3ce80ef2-925e-4a36-b6d1-5475868714eb --limit 1 --output json`
- `make format`
- `make check-command-docs`
- `make lint`

Closes #1292.

### DIFF
--- a/internal/asc/client.go
+++ b/internal/asc/client.go
@@ -814,8 +814,14 @@ func (c *Client) GetBetaTesters(ctx context.Context, appID string, opts ...BetaT
 			return nil, fmt.Errorf("betaTesters: %w", err)
 		}
 		path = query.nextURL
-	} else if queryString := buildBetaTestersQuery(appID, query); queryString != "" {
-		path += "?" + queryString
+	} else {
+		queryString, err := buildBetaTestersQuery(appID, query)
+		if err != nil {
+			return nil, fmt.Errorf("betaTesters: %w", err)
+		}
+		if queryString != "" {
+			path += "?" + queryString
+		}
 	}
 
 	data, err := c.do(ctx, "GET", path, nil)
--- a/internal/asc/client_http_test.go
+++ b/internal/asc/client_http_test.go
@@ -2023,7 +2023,7 @@ func TestGetBetaTesters_WithAppFilter(t *testing.T) {
 }
 
 func TestGetBetaTesters_WithBuildFilter(t *testing.T) {
-	// API only allows one relationship filter, so builds takes precedence over apps
+	// API only allows one relationship filter; build filter takes precedence over apps.
 	response := jsonResponse(http.StatusOK, `{"data":[{"type":"betaTesters","id":"1","attributes":{"email":"tester@example.com"}}]}`)
 	client := newTestClient(t, func(req *http.Request) {
 		if req.Method != http.MethodGet {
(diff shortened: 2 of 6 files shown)


rudrankriyam:
Addressed review comment on `buildBetaTestersQuery` (lines 806-808):

**Fix**: `buildBetaTestersQuery` now returns an error when both `filterBuilds` and `groupIDs` are set, so every call site gets a consistent rejection instead of silently dropping the group filter. Also added CLI-level `--group` + `--build-id` validation to `beta-testers export` (matching the existing check in `beta-testers list`).

**Tests added**:
- `TestBuildBetaTestersQueryRejectsGroupAndBuild` (unit)
- `TestGetBetaTesters_RejectsGroupAndBuildConflict` (HTTP integration)
- `TestTestFlightBetaTestersExportRejectsGroupAndBuildID` (cmdtest)

**Commands run**: `make format`, `make lint`, `ASC_BYPASS_KEYCHAIN=1 make test` -- all pass.

PR #1321, opened 2026-04-01
[codex] Fix ASC Studio app selection races

rudrankriyam:
## Summary
- make ASC Studio app switching lightweight by loading only overview metadata on selection and moving app-scoped section fetches behind the active section
- reset per-app section and TestFlight state on app changes so stale data cannot bleed into the next selection
- add a shell-level error boundary and null-safe version badge rendering, plus rebuild the tracked Studio frontend bundle

## Root cause
Switching apps from the sidebar mixed the old active section with the new selection request and also fanned out a large burst of app-scoped subprocesses immediately on click. That combination made it possible for stale app data to race with the new selection and increased the odds of the shell blanking when a render-time error escaped the inner content boundary.

## Impact
- app picker clicks now return the shell to `overview` deterministically before section-specific loading starts
- expensive app-scoped data fetches only happen when the user opens that section
- missing version state values no longer crash sidebar rendering
- shell-level render failures now show a fallback instead of leaving the Wails webview visually blank

## Validation
- `make studio-frontend-test`
- `make studio-frontend-build`
- `make format`
- `make check-command-docs`
- `make lint`


rudrankriyam:
Second pass is now on this branch in `ba3c7d55`.

This follow-up closes the remaining Studio gaps from the deeper audit:
- make app selection synchronous instead of wrapping the reset in `startTransition`
- refresh bootstrap/auth/app list even when an app is already selected, then reload the visible section
- retry standalone sections after cached transient errors instead of treating an error entry as a permanent cache hit
- harden `TestFlightView`, `SubscriptionsView`, and `PricingView` against missing/null status/date fields
- add regression coverage for standalone retry, refresh-with-selected-app, and null-safe section rendering

Validation rerun on this branch:
- `make studio-frontend-test`
- `make studio-frontend-build`
- `make format`
- `make check-command-docs`
- `make lint`

PR #1384, opened 2026-04-06
[codex] make asc validate include remediation by default

### LINKED ISSUE #1364: Feature: add validate remediation modes with --next and --fix-plan
## Summary

add validate remediation modes with --next and --fix-plan

## Reproduction

```bash
asc validate --app "APP_ID" --version "1.2.3"
```

## Expected behavior

Validate should be able to emit an ordered remediation summary such as --next or --fix-plan instead of only reporting blockers.

## Actual behavior

```
The current validate flow reports blocking issues but does not produce a next-step or fix-plan mode.
```

## Environment

- **asc version:** dev (commit: unknown, date: unknown)
- **OS:** darwin/arm64
- **Filed via:** `asc snitch`

rudrankriyam:
## Summary
- make `asc validate` include an ordered remediation plan in the default report
- keep `asc validate` as one canonical readiness command instead of splitting remediation into alternate modes
- render remediation across JSON, table, and markdown while preserving blocking exit behavior
- reject misplaced top-level `validate` flags before subcommands instead of silently ignoring them

## Problem
`asc validate` already knew what was wrong, but it did not make the next action obvious. The issue originally proposed `--next` and `--fix-plan`, but that would have required callers to pick a separate view before the CLI answered the core workflow question: "am I ready, and if not, what should I fix first?"

## Direction Change
The implementation intentionally changed direction from the issue wording.

Instead of introducing separate remediation modes, this PR makes remediation part of the main `asc validate` response. That keeps the command aligned with the rest of the CLI: one canonical validation command, one source of truth, one output that answers both readiness and next steps.

## First-Principles Design
- `asc validate` is the canonical pre-submission readiness command, so a single invocation should answer readiness and remediation together.
- Remediation should be derived from the same check set as the report so there is one source of truth, not parallel output paths.
- The default output should stay audit-friendly: summary first, ordered remediation next, detailed checks after that.
- The CLI should fail clearly on ambiguous flag placement rather than accept flags that get ignored.

## What Changed
- `validation.Report` now includes a `remediation` section with ordered actionable steps.
- Remediation steps are ranked by blocking severity, with stable ordering preserved within each severity bucket.
- `asc validate` table and markdown output now show summary, remediation, and then raw checks.
- JSON output remains the same shape plus one additive `remediation` object.
- The separate `--next` and `--fix-plan` flags were removed in favor of the default report carrying its own remediation.
- `asc validate` now rejects parent flags placed before `testflight`, `iap`, or `subscriptions` when those flags would otherwise be ignored.

## Why This Shape Fits The CLI
- Users should not need to learn alternate modes to get the next fix. The first remediation step is the next fix.
- Keeping one report avoids duplicate docs, duplicated tests, and divergent behavior between "report" and "plan" paths.
- The detailed checks still remain available, so the command stays useful for CI, review, and debugging instead of becoming an opaque recommendation engine.

## Alternatives Considered
- Keep `--next` and `--fix-plan` as separate flags.
This makes the feature explicit, but it fragments the main workflow and adds extra surface area for a command that should stay canonical.
- Replace the detailed check list with remediation only.
This would simplify the output, but it would remove the evidence users need to understand why a step is recommended and would make machine and human auditing worse.

## Compatibility And Blast Radius
- No new stable top-level flags were added.
- Default `asc validate` exit behavior for blocking issues is unchanged.
- JSON output is additive through the new `remediation` field.
- Table and markdown output become richer by default because remediation is now included.
- Removed experimental remediation flags now return usage error `2` instead of continuing on an obsolete path.
- The stricter subcommand flag parsing is an intentional safety improvement and only affects previously ambiguous invocations.

## Expected Usage
- `asc validate --app "APP_ID" --version-id "VERSION_ID"`
- `asc validate --app "APP_ID" --version "1.0.0" --platform IOS`
- `asc validate --app "APP_ID" --version-id "VERSION_ID" --strict`

A successful JSON response now carries `summary`, `remediation`, and `checks` together, so callers can both act on the next step and inspect the full evidence set from one invocation.

## Validation
- `go run . validate --help`
- `make format`
- `make check-command-docs`
- `make lint`
- `make build`

Closes #1364.

### DIFF
--- a/internal/asc/output_validate.go
+++ b/internal/asc/output_validate.go
@@ -11,15 +11,11 @@ func init() {
 	registerDirect(func(v *validation.Report, render func([]string, [][]string)) error {
 		h, r := validationSummaryRows(v)
 		render(h, r)
+		if len(v.Remediation.Steps) > 0 {
+			rh, rr := validationRemediationRows(v)
+			render(rh, rr)
+		}
 		oh, or := validationCheckRows(v)
-		render(oh, or)
-		return nil
-	})
-
-	registerDirect(func(v *validation.RemediationReport, render func([]string, [][]string)) error {
-		h, r := remediationSummaryRows(v)
-		render(h, r)
-		oh, or := remediationStepRows(v)
 		render(oh, or)
 		return nil
 	})
@@ -52,7 +48,7 @@ func init() {
 }
 
 func validationSummaryRows(report *validation.Report) ([]string, [][]string) {
-	headers := []string{"App ID", "Version ID", "Version", "Platform", "Errors", "Warnings", "Infos", "Blocking", "Strict"}
+	headers := []string{"App ID", "Version ID", "Version", "Platform", "Errors", "Warnings", "Infos", "Blocking", "Actionable", "Strict"}
 	rows := [][]string{{
 		report.AppID,
 		report.VersionID,
@@ -62,6 +58,7 @@ func validationSummaryRows(report *validation.Report) ([]string, [][]string) {
 		fmt.Sprintf("%d", report.Summary.Warnings),
 		fmt.Sprintf("%d", report.Summary.Infos),
 		fmt.Sprintf("%d", report.Summary.Blocking),
+		fmt.Sprintf("%d", report.Remediation.TotalActionable),
 		formatBool(report.Strict),
 	}}
 	return headers, rows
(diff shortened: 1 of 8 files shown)


rudrankriyam:
@codex please review this PR.

PR #1435, opened 2026-04-14
[codex] app-clips: support inline invocation localizations

### LINKED ISSUE #1434: app-clips invocations create fails: no way to provide required inline localizations
## Summary

`asc app-clips invocations create` always fails because the App Store Connect API requires the `betaAppClipInvocationLocalizations` relationship to be provided inline, but the CLI has no way to supply it.

## Repro

```bash
asc app-clips invocations create --build-bundle-id "BUILD_BUNDLE_ID" --url "https://example.com/clip"
```

## Actual

```
Error: The provided entity is missing a required relationship:
You must provide a value for the relationship 'betaAppClipInvocationLocalizations' with this request
```

The `--localization-id` flag exists but requires pre-existing localization IDs. However, localizations can only be created via `asc app-clips invocations localizations create --invocation-id ...`, which requires an invocation to already exist — a chicken-and-egg problem.

## Expected

Either:
1. Add `--locale` and `--title` flags to `invocations create` that get sent as inline `included[]` localizations in the POST body, or
2. Auto-create a default localization when none is provided

## Workaround

Had to use raw `curl` against the API with an inline `included[]` array:

```bash
TOKEN=$(asc auth token --confirm)
curl -X POST "https://api.appstoreconnect.apple.com/v1/betaAppClipInvocations" \
  -H "Authorization: Bearer $TOKEN" \
  -H "Content-Type: application/json" \
  -d '{
    "data": {
      "type": "betaAppClipInvocations",
      "attributes": { "url": "https://example.com/clip" },
      "relationships": {
        "buildBundle": { "data": { "type": "buildBundles", "id": "BUILD_BUNDLE_ID" } },
        "betaAppClipInvocationLocalizations": {
          "data": [{ "type": "betaAppClipInvocationLocalizations", "id": "${new-loc}" }]
        }
      }
    },
    "included": [{
      "type": "betaAppClipInvocationLocalizations",
      "id": "${new-loc}",
      "attributes": { "title": "My Clip", "locale": "en-AU" }
    }]
  }'
```

## Environment

- asc version: 1.2.1
- OS: macOS arm64

rudrankriyam:
## Summary
- add `--locale` and `--title` to `asc app-clips invocations create` so the command can send inline `included[]` localizations
- keep `--localization-id` working for pre-existing localizations and validate locally that at least one localization source is provided
- cover the client payload, CLI validation, and both create paths with focused tests

## Why
Fixes the chicken-and-egg flow in #1434: the App Store Connect API requires `betaAppClipInvocationLocalizations` on create, but the CLI could only reference existing localization IDs.

## Approach
This PR chooses explicit inline-localization flags instead of auto-creating a default localization because:
- it matches the issue’s preferred UX directly
- it keeps locale/title explicit and reviewable
- it avoids hidden defaults in a mutating command

Alternatives considered:
- auto-create a default localization when none is provided: simpler CLI surface, but it would require guessing locale/title semantics
- leave validation to the API: smaller change, but it preserves the current broken UX and poor error loop

## Validation
- `make format`
- `make check-command-docs`
- `make lint`
- `go build -o /tmp/asc .`
- built binary check: `/tmp/asc app-clips invocations create --build-bundle-id bundle-1 --url https://example.com/clip` exits `2` with `Error: provide --localization-id or both --locale and --title`

## Live ASC verification
- Verified read-only discovery with `ASC_BYPASS_KEYCHAIN=1 /tmp/asc builds list --app 6759231657 --limit 5 --output json` and confirmed the preferred disposable app currently has zero builds, so there was no safe throwaway build bundle to mutate.
- Confirmed read-only build-bundle and invocation listing on another app, but intentionally did not create/delete an invocation on a non-throwaway app.

Closes #1434

### DIFF
--- a/internal/asc/client_app_clips.go
+++ b/internal/asc/client_app_clips.go
@@ -420,6 +420,14 @@ type BetaAppClipInvocationLocalizationCreateAttributes struct {
 	Locale string `json:"locale"`
 }
 
+// BetaAppClipInvocationLocalizationInlineCreate describes an inline localization create payload.
+type BetaAppClipInvocationLocalizationInlineCreate struct {
+	Type          ResourceType                                      `json:"type"`
+	ID            string                                            `json:"id,omitempty"`
+	Attributes    BetaAppClipInvocationLocalizationCreateAttributes `json:"attributes"`
+	Relationships *BetaAppClipInvocationLocalizationRelationships   `json:"relationships,omitempty"`
+}
+
 // BetaAppClipInvocationLocalizationUpdateAttributes describes localization update attributes.
 type BetaAppClipInvocationLocalizationUpdateAttributes struct {
 	Title *string `json:"title,omitempty"`
@@ -445,7 +453,8 @@ type BetaAppClipInvocationCreateData struct {
 
 // BetaAppClipInvocationCreateRequest is the create request payload.
 type BetaAppClipInvocationCreateRequest struct {
-	Data BetaAppClipInvocationCreateData `json:"data"`
+	Data     BetaAppClipInvocationCreateData                 `json:"data"`
+	Included []BetaAppClipInvocationLocalizationInlineCreate `json:"included,omitempty"`
 }
 
 // BetaAppClipInvocationUpdateData is the payload for updating an invocation.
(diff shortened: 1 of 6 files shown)


rudrankriyam:
@klemenkosir see

klemenkosir:
image

PR #1492, opened 2026-05-06
Add web subscription remove from sale

mahmoudashraf93:
## Summary

Adds an experimental web-session command for removing an approved auto-renewable subscription from sale:

```bash
asc web subscriptions availability remove-from-sale \
  --app "APP_ID" \
  --subscription-id "PRODUCT_ID_OR_SUBSCRIPTION_ID" \
  --confirm
```

The command follows the App Store Connect web flow we observed for this action by patching the subscription plan availability, setting `availableInNewTerritories=false`, and clearing `availableTerritories`.

## Why

The public App Store Connect API does not expose a documented first-class operation for removing an approved auto-renewable subscription from sale. App Store Connect performs this through the internal web `/iris` plan availability endpoint, and this repo already has an experimental `asc web` command family for reviewer-owned web-session workflows where the public API falls short.

This keeps the feature in the web-session namespace instead of presenting it as a stable public API command.

## Behavior

- Adds `asc web subscriptions availability remove-from-sale`.
- Requires `--confirm` because this is a destructive sale-availability mutation.
- Accepts `--subscription-id` as a subscription ID, product ID, or exact current name.
- Accepts `--app` to resolve product ID/name through existing web subscription listing.
- Accepts `--plan-availability-id` for direct mutation when the internal availability ID is already known.
- Returns structured JSON/table/markdown output with the selected plan availability and removal state.
- Converts Apple `403` responses into an explicit Account Holder role error, matching App Store Connect behavior for approved subscriptions.

## Validation

Local gates:

```bash
make format
make check-command-docs
make lint
ASC_BYPASS_KEYCHAIN=1 make test
go build -o /tmp/asc-subscription-remove-from-sale .
```

Black-box CLI checks:

```bash
/tmp/asc-subscription-remove-from-sale web subscriptions availability remove-from-sale --help
/tmp/asc-subscription-remove-from-sale web subscriptions availability remove-from-sale --subscription-id SUB_ID
```

Live App Store Connect smoke test against a user-approved subscription:

```bash
/tmp/asc-subscription-remove-from-sale web review subscriptions list \
  --apple-id "REDACTED" \
  --app "REDACTED_APP_ID" \
  --output json \
  --pretty

/tmp/asc-subscription-remove-from-sale web subscriptions availability remove-from-sale \
  --apple-id "REDACTED" \
  --app "REDACTED_APP_ID" \
  --subscription-id "REDACTED_PRODUCT_ID" \
  --confirm \
  --output json \
  --pretty
```

Result:

```json
{
  "subscriptionId": "REDACTED_SUBSCRIPTION_ID",
  "planAvailabilityId": "REDACTED_PLAN_AVAILABILITY_ID",
  "planType": "UPFRONT",
  "removedFromSale": true,
  "availableInNewTerritories": false,
  "availableTerritories": null,
  "requiresAccountHolderRole": true
}
```

## Notes

This is intentionally marked experimental and documented under `asc web` because it depends on private, undocumented Apple web-session endpoints.

## Summary by CodeRabbit

* **New Features**
  * New "asc web subscriptions availability remove-from-sale" command available under the web command to remove a subscription plan from sale; requires Account Holder session and explicit confirmation.

* **Documentation**
  * Added docs with usage examples, Account Holder session warning, and guidance for supplying a plan-availability-id.

* **Tests**
  * Added comprehensive tests covering removal flows, ID validation, verification/read-back, authorization guidance, progress messaging, and error cases.

### DIFF
--- a/commands/web.mdx
+++ b/commands/web.mdx
@@ -21,6 +21,9 @@ asc web apps create --name "My App" --bundle-id "com.example.app" --sku "MYAPP12
 # Bootstrap app availability where the public API falls short
 asc web apps availability create --app "123456789" --territory "USA,GBR" --available-in-new-territories false
 
+# Remove an approved auto-renewable subscription from sale
+asc web subscriptions availability remove-from-sale --subscription-id "SUB_ID" --confirm
+
 # Declare that an app is not a regulated medical device
 asc web apps medical-device set --app "123456789" --declared false
 
@@ -125,6 +128,35 @@ Available commands:
 * `asc web review list` - list review submissions for an app
 * `asc web review show` - inspect one submission and related messages/screenshots
 * `asc web review subscriptions` - inspect and mutate next-version subscription review selection
+
+### `asc web subscriptions`
+
+Subscription sale availability workflows over Apple web-session `/iris` endpoints.
+
+Available commands:
+
+* `asc web subscriptions availability remove-from-sale` - remove an approved auto-renewable subscription from sale
+
+<Warning>
+  Removing an approved auto-renewable subscription from sale requires an Account Holder web session. App Store Connect rejects Admin and App Manager users for this action.
+</Warning>
+
+Example:
+
+```bash  theme={null}
+asc web subscriptions availability remove-from-sale \
+  --subscription-id "SUB_ID" \
+  --confirm
+```
+
+If you know the internal plan availability ID, pass it directly:
+
+```bash  theme={null}
+asc web subscriptions availability remove-from-sale \
+  --subscription-id "SUB_ID" \
+  --plan-availability-id "PLAN_AVAILABILITY_ID" \
+  --confirm
+```
 
 ### `asc web analytics`
 
--- a/internal/cli/cmdtest/web_subscriptions_test.go
+++ b/internal/cli/cmdtest/web_subscriptions_test.go
@@ -1,0 +1,191 @@
+package cmdtest
+
+import (
+	"context"
+	"encoding/json"
+	"io"
+	"net/http"
+	"strings"
+	"testing"
+
+	cmd "github.com/rudrankriyam/App-Store-Connect-CLI/cmd"
+	webcmd "github.com/rudrankriyam/App-Store-Connect-CLI/internal/cli/web"
+	webcore "github.com/rudrankriyam/App-Store-Connect-CLI/internal/web"
+)
+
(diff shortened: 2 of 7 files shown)


rudrankriyam:
@mithileshchellappan can you review?

mithileshchellappan:
I reviewed the new `asc web subscriptions availability remove-from-sale` flow and found two blocking issues that should be fixed before merge.

1. **Verify direct plan IDs before patching.** When `--plan-availability-id` is supplied, the command skips the subscription-scoped availability lookup and PATCHes that plan ID directly (`internal/cli/web/web_subscriptions.go:149-165`). That means a stale or mistyped plan availability ID can remove a different subscription plan while the result still reports the user-provided `subscriptionId`. For a destructive command, please verify the plan belongs to the resolved subscription/app before mutation, or make direct plan-ID mode explicitly plan-scoped and avoid claiming a subscription association.

2. **Do not report removal without verifying response state.** After any successful PATCH, the command hardcodes `RemovedFromSale: true` (`internal/cli/web/web_subscriptions.go:173-179`). It does not fail if Apple returns `availableInNewTerritories=true`, non-empty `availableTerritories`, or a partial response that omits the relationship. Since this command advertises removal from sale, it should derive success from the response and ideally perform a readback/follow-up verification before printing `removedFromSale=true`.

I also compared this against the existing public availability surface. The public API exposes `POST /v1/subscriptionAvailabilities` plus read endpoints, and this repo already has `asc subscriptions pricing availability edit`, but that path requires non-empty territories and does not model clearing sale availability entirely. So the web command may still be justified, but these destructive success/target checks need to be tightened.

mahmoudashraf93:
Thanks @mithileshchellappan addressing those right away

mahmoudashraf93:
Addressed both blockers Direct `--plan-availability-id` mode now still loads the subscription-scoped plan availability list first and refuses to PATCH if the supplied plan ID is not attached to the resolved subscription.

The command no longer hardcodes removal success from a 2xx PATCH. It performs a post-PATCH readback of the subscription plan availability and only prints `removedFromSale: true` when `availableInNewTerritories=false`, the available territories relationship was loaded, and the territory list is empty.

rudrankriyam (approved):
Audited hard. Verified the destructive remove-from-sale path is scoped through subscription-owned availability lookup, reads back the post-patch state, and fails if removal is not confirmed. Full local guardrails and GitHub checks are green.

PR #1571, opened 2026-05-26
Fix iris IAP listing fields + accept productId for --iap-id

Max-Bin:
## Summary

`asc web review iaps attach` fails on real apps for two compounding reasons. This PR fixes both, plus a correctness issue surfaced in review.

### Bug 1 — iris listing rejects the requested fields

`FindReviewIAP` requests `fields[inAppPurchases]=productId,name,state,inAppPurchaseType,isAppStoreReviewInProgress,submitWithNextAppStoreVersion` and `sort=name`. Apple's iris flavor of `/iris/v1/apps/{APP_ID}/inAppPurchases` only accepts `productId,referenceName,state` — `name`, `inAppPurchaseType`, `isAppStoreReviewInProgress`, and `submitWithNextAppStoreVersion` all return `PARAMETER_ERROR.INVALID` (HTTP 400), so every attach call fails before the resource match runs.

### Bug 2 — `FindReviewIAP` matches only the opaque iris resource id

Even after the listing returns 200, the post-`a043a87` ("Harden web review IAP attach") match predicate `strings.TrimSpace(resource.ID) == iapID` never hits in production: the iris listing returns each IAP with a UUID-shaped resource id distinct from the numeric ID surfaced by the public ASC REST API.

```
$ asc web review iaps attach --app APP_ID --iap-id IAP_ID --confirm
Error: web review iaps attach: in-app purchase "IAP_ID" was not found
under app "APP_ID"; refusing to attach
```

### Bug 3 — attach POST sent the caller's selector instead of the resolved iris id

When `FindReviewIAP` matched via `attributes.productId`, the attach POST was still sending the original `--iap-id` selector as the relationship id. Posting a bundle-style productId where iris expects its own resource id is incorrect. The fix is to thread `reviewIAP.ID` (the resolved iris UUID) through both the idempotent short-circuit and the `CreateInAppPurchaseSubmission` call.

## Fix

1. **Listing fields** trimmed to `productId,referenceName,state` and `sort=referenceName`. The struct gains `ReferenceName` (replacing the iris-unavailable `Name`) and drops `InAppPurchaseType` / `IsAppStoreReviewInProgress`. `SubmitWithNextAppStoreVersion` stays on the struct (the `/inAppPurchaseSubmissions` POST still carries it; idempotency short-circuit works against that response).
2. **Matcher** now resolves `decoded.ID == iapID || decoded.ProductID == iapID`, mirroring how `findReviewSubscription` already does ID+productID resolution. The wrong-app safety guard still holds — a different app's IAP won't have a matching productId either.
3. **Attach POST** now uses the resolved iris UUID (`reviewIAP.ID`) as the relationship id, not the caller's selector. A new test (`TestWebReviewIAPsAttachPostsIrisResourceIDWhenSelectorWasProductID`) asserts the POST body carries the iris UUID and does not carry the productId selector.
4. **Validator** relaxed to accept both numeric and bundle-style `--iap-id`. The iris matcher then resolves whichever form the iris response actually carries.

## Contract: which `--iap-id` forms are in scope

| `--iap-id` passed              | Matches against           | Supported in this PR? |
| ------------------------------ | ------------------------- | --------------------- |
| iris resource UUID             | `resource.ID`             | ✅                     |
| bundle-style `com.example.…`   | `attributes.productId`    | ✅                     |
| public numeric ASC IAP id      | (not exposed by iris)     | ❌ — out of scope      |

iris does not surface the public numeric ASC IAP id (the value `asc iap list` outputs / the ASC URL bar shows). Callers with only the public numeric id need to look it up via `asc iap list`, copy the `productId` field, then pass that to `--iap-id`. A `numericID → productId` bridge could be added as a follow-up PR if the explicit lookup step turns out to be a meaningful UX gap.

The pre-existing `TestFindReviewIAPMatchesByProductID` was misnamed — its selector happened to match the resource id, not `productId`, so it kept passing while the production code path was broken. Updated to actually exercise the fallback with a UUID-shaped resource id.

## Test plan

- [x] `go test ./internal/web/... ./internal/cli/web/... ./internal/cli/cmdtest/...` — all green; new tests guard against re-adding any of the four iris-rejected field names, and against the attach POST sending the caller's selector instead of the iris UUID.
- [x] `go vet ./...` clean.
- [x] `go build ./...` clean.
- [x] Verified end-to-end on a live app:
  - Without this patch: iris listing returns HTTP 400 (`PARAMETER_ERROR.INVALID`) on the first request → attach aborts.
  - With this patch: iris listing returns HTTP 200; `FindReviewIAP` matches by productId; POST `/inAppPurchaseSubmissions` proceeds and flips `submitWithNextAppStoreVersion: true` (or returns 409 if the IAP was already attached on a prior run — the expected idempotency response).

## Notes

Follow-up to PR #1570 (which I authored). The a043a87 hardening intent was good — fail fast if the user passes an IAP belonging to a different app — but the matching predicate was too strict and the field list was broader than what iris accepts. With both fixes plus the POST-id correctness commit, the safety guard still holds while the legitimate path becomes usable.

## Summary by CodeRabbit

* **Bug Fixes**
  * IAP review listing/matching now accepts productId-style selectors, matches by selector or backend resource ID, uses the resolved backend ID when creating submissions, and sorts/page-queries by reference name. Improved detection of "already exists" conflict responses.

* **Tests**
  * Added/updated tests for productId matching, relaxed IAP-ID validation, attach-flow behavior (preserving caller selector while using resolved IDs), and conflict-detection handling.

[Review Change Stack](https://app.coderabbit.ai/change-stack/rorkai/App-Store-Connect-CLI/pull/1571?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/internal/cli/cmdtest/web_review_iaps_test.go
+++ b/internal/cli/cmdtest/web_review_iaps_test.go
@@ -306,6 +306,11 @@ func TestWebReviewIAPsAttachInvalidValueExitCodes(t *testing.T) {
 			wantStderr: `invalid boolean value "maybe" for -confirm`,
 		},
 		{
+			name:       "blank iap selector",
+			args:       []string{"web", "review", "iaps", "attach", "--app", "123456789", "--iap-id", "   ", "--confirm"},
+			wantStderr: "--iap-id is required",
+		},
+		{
 			name:       "subcommand flag before attach rejected",
 			args:       []string{"web", "review", "iaps", "--app", "123456789", "attach", "--iap-id", "9000000001", "--confirm"},
 			wantStderr: "flag provided but not defined: -app",
--- a/internal/cli/cmdtest/web_review_iaps_test.go
+++ b/internal/cli/cmdtest/web_review_iaps_test.go
@@ -174,11 +174,6 @@ func TestWebReviewIAPsAttachValidationErrors(t *testing.T) {
 			args:       []string{"web", "review", "iaps", "attach", "--app", "com.example.app", "--iap-id", "9000000001", "--confirm"},
 			wantStderr: "--app must be a numeric App Store Connect app ID",
 		},
-		{
-			name:       "non numeric iap",
-			args:       []string{"web", "review", "iaps", "attach", "--app", "123456789", "--iap-id", "com.example.pro", "--confirm"},
-			wantStderr: "--iap-id must be a numeric App Store Connect in-app purchase ID",
-		},
 	}
 
 	for _, test := range tests {
@@ -284,11 +279,6 @@ func TestWebReviewIAPsAttachInvalidValueExitCodes(t *testing.T) {
 			name:       "invalid app selector",
 			args:       []string{"web", "review", "iaps", "attach", "--app", "com.example.app", "--iap-id", "9000000001", "--confirm"},
 			wantStderr: "--app must be a numeric App Store Connect app ID",
-		},
-		{
-			name:       "invalid iap selector",
-			args:       []string{"web", "review", "iaps", "attach", "--app", "123456789", "--iap-id", "com.example.pro", "--confirm"},
-			wantStderr: "--iap-id must be a numeric App Store Connect in-app purchase ID",
 		},
 		{
 			name:       "invalid confirm value",
--- a/internal/cli/cmdtest/web_review_iaps_test.go
+++ b/internal/cli/cmdtest/web_review_iaps_test.go
@@ -21,6 +21,11 @@ import (
 func TestWebReviewIAPsAttachRootSuccess(t *testing.T) {
 	setupCachedWebReviewIAPSession(t, "user@example.com")
 
+	const (
+		irisResourceID = "11111111-2222-3333-4444-555555555555"
+		productID      = "com.example.removeads"
+	)
+
 	requests := newRequestLog(3)
 	installDefaultTransport(t, roundTripFunc(func(req *http.Request) (*http.Response, error) {
 		requests.Add(req.Method + " " + req.URL.Host + req.URL.Path)
@@ -45,17 +50,16 @@ func TestWebReviewIAPsAttachRootSuccess(t *testing.T) {
 				t.Fatalf("expected web IAP lookup limit=300, got %q", got)
 			}
 			return webReviewIAPJSONResponse(http.StatusOK, `{
-				"data": [{
-					"type": "inAppPurchases",
-					"id": "9000000001",
-					"attributes": {
-						"name": "Remove Ads",
-						"productId": "com.example.removeads",
-						"inAppPurchaseType": "NON_CONSUMABLE",
-						"state": "READY_TO_SUBMIT"
-					}
-				}]
-			}`, req), nil
+					"data": [{
+						"type": "inAppPurchases",
+						"id": "`+irisResourceID+`",
+						"attributes": {
+							"productId": "`+productID+`",
+							"referenceName": "Remove Ads",
+							"state": "READY_TO_SUBMIT"
+						}
+					}]
+				}`, req), nil
 		case req.Method == http.MethodPost &&
 			req.URL.Host == "appstoreconnect.apple.com" &&
 			req.URL.Path == "/iris/v1/inAppPurchaseSubmissions":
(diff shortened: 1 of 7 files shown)


rudrankriyam:
Yikes, fixing

rudrankriyam:
@Max-Bin have you tested this PR on a live app and it works fine?

rudrankriyam:
Thanks for digging into this, and I believe you found a real regression from the hardening in #1570.

Could you share one redacted example of the identifiers you saw for the same IAP across the two surfaces?

```json
// public ASC API / asc iap list
{"id":"PUBLIC_ASC_IAP_ID","attributes":{"productId":"PRODUCT_ID"}}

// iris /apps/{appID}/inAppPurchases
{"id":"IRIS_RESOURCE_ID","attributes":{"productId":"IRIS_PRODUCT_ID"}}
```

The bit I want to confirm is whether the value passed to `--iap-id` matched `IRIS_PRODUCT_ID`, or whether it was the public numeric ASC IAP id. Today `asc web review iaps attach` validates `--iap-id` as numeric, so if callers pass the public numeric id and Iris only exposes the bundle-style `productId`, we may need to resolve public numeric id -> productId first, then attach using the Iris resource id. If your debug showed Iris `attributes.productId` carrying the numeric id, this PR is the right shape and we can update the tests/description around that exact behavior.

Max-Bin:
Yes — tested on a live production app.

With current `main`: the iris listing call (the very first request `FindReviewIAP` makes) returns HTTP 400 `PARAMETER_ERROR.INVALID` because four of the requested `fields[inAppPurchases]` keys (`name`, `inAppPurchaseType`, `isAppStoreReviewInProgress`, `submitWithNextAppStoreVersion`) aren't accepted on the iris flavor of that resource — iris uses `referenceName`. Attach never gets past the listing.

With this PR: iris listing returns HTTP 200, `FindReviewIAP` matches on the iris `productId` attribute, and `POST /inAppPurchaseSubmissions` proceeds — flipping `submitWithNextAppStoreVersion: true` on a fresh IAP, or returning HTTP 409 cleanly on a re-run for an already-attached IAP (the expected idempotency response).

Max-Bin:
Good question — here's what I actually saw, redacted to placeholders.

**Public ASC API (`asc iap list --app APP_ID`):**
```json
{"id":"PUBLIC_ASC_IAP_ID","attributes":{
  "name":"Pro Lifetime",
  "productId":"com.example.app.pro.lifetime",
  "inAppPurchaseType":"NON_CONSUMABLE",
  "state":"READY_TO_SUBMIT"
}}
```
- `id` is a numeric ASC IAP id (the value `asc iap list` outputs and the one the ASC URL bar shows).
- `productId` is the bundle-style identifier from the StoreKit config.

**iris `GET /iris/v1/apps/{APP_ID}/inAppPurchases`** (with `fields[inAppPurchases]=productId,referenceName,state`):
```json
{"id":"IRIS_RESOURCE_UUID","attributes":{
  "productId":"com.example.app.pro.lifetime",
  "referenceName":"Pro Lifetime",
  "state":"READY_TO_SUBMIT"
}}
```
- `id` is a UUID-shaped opaque resource id, **distinct** from `PUBLIC_ASC_IAP_ID`.
- `productId` is the **same** bundle-style identifier as the public surface (so `productId` is the only field stable across both).
- The public numeric ASC IAP id does **not** appear anywhere in the iris response.

So to answer the direct question: I passed the bundle-style productId (`com.example.app.pro.lifetime`) as `--iap-id`, and `FindReviewIAP` matched on `attributes.productId`. iris's `attributes.productId` was the bundle-style identifier, not the public numeric id.

**Three caller flavors and what this PR supports**

| `--iap-id` passed              | Matches against                       | Works in this PR? |
| ------------------------------ | ------------------------------------- | ----------------- |
| iris resource UUID             | `resource.ID`                         | ✅ (already did)   |
| bundle-style `com.example.…`   | `attributes.productId`                | ✅ (this PR)       |
| public numeric ASC IAP id      | (nothing in iris matches it)          | ❌                 |

The third row would need a separate `numericID → productId` bridge (one public REST call before the iris matcher) — but I'd suggest leaving that out of this PR to keep scope tight and the semantics aligned with `findReviewSubscription`, which also doesn't bridge. With the validator relaxed, the workflow becomes "copy `productId` from `asc iap list` output → pass to `--iap-id`," which is one extra step but works today. Happy to file a follow-up PR for the public-numeric-id bridge if you decide it's worth supporting.

rudrankriyam:
Thanks, this is super helpful. The redacted mapping makes the behavior much clearer, and the Iris field-list finding looks like the real blocker on `main`.

Given what you saw, I think the PR shape should be:

- support Iris UUID and bundle-style `productId` in `--iap-id`
- explicitly not claim support for the public numeric ASC IAP id in this PR, since Iris does not expose it
- potentially leave public numeric id -> productId bridging as a follow-up

One remaining code concern: when `FindReviewIAP` matches by `productId`, can you make the attach mutation use the matched Iris resource id (`reviewIAP.ID`) rather than the original `--iap-id` selector? Otherwise `--iap-id com.example.app.pro.lifetime` can pass verification but still POST the productId as the relationship id instead of the Iris UUID.

After that, could you also tweak the PR description to remove/soften the parts that say the public numeric id resolves cleanly here? The table in your comment is the clearest contract: Iris UUID works, productId works, public numeric id is a separate bridge.

Max-Bin:
Addressed both points — thanks for catching the POST-id mismatch.

**Code change (new commit 24d84ce):** `FindReviewIAP`'s result is now threaded through both the idempotent short-circuit and the attach POST as `reviewIAP.ID`, so when `--iap-id` was a bundle-style productId the POST still carries the iris UUID as the `inAppPurchaseV2` relationship id. Added `TestWebReviewIAPsAttachPostsIrisResourceIDWhenSelectorWasProductID` to lock this in — it asserts the POST body contains the iris UUID and does **not** contain the productId selector.

**PR description:** softened the wording around `--iap-id` shapes to match the table — iris UUID and bundle-style productId are explicitly in scope; public numeric ASC IAP id is explicitly out of scope, with a note that callers can look up `productId` via `asc iap list` as the workaround and that a `numericID → productId` bridge could be a follow-up.

Let me know if you'd like the validator comment / `--iap-id` flag help text tightened further along the same lines.

rudrankriyam:
Audit follow-up pushed.

I added three small hardening commits on top of Max’s fix:

- `237b82e3` handles the already-attached `ALREADY_EXISTS` 409 as `changed=false`, preserving idempotent reruns now that the Iris list fields are sparse.
- `09917166` makes the cmdtest success path use a bundle-style `--iap-id` and verifies the POST relationship uses the resolved Iris UUID.
- `490ef3fe` narrows the idempotent 409 detection so broader conflicts like “already attached to another submission” still surface as errors, plus adds whitespace `--iap-id` usage coverage.

Local verification passed:

- `ASC_BYPASS_KEYCHAIN=1 go run . web review iaps attach --help`
- `make format`
- `make check-command-docs`
- `make lint`
- focused `internal/web`, `internal/cli/web`, and `internal/cli/cmdtest` tests for this flow

I resolved the review threads that these commits address.

rudrankriyam:
@coderabbitai review

rudrankriyam (dismissed):
Hard audit looks good now. All review threads are resolved, CodeRabbit is clean, CI is green, and the branch is current with main. The PR scope is clear: Iris UUID and bundle-style productId are supported; public numeric ASC IAP id bridging can be a follow-up.

rudrankriyam (dismissed):
Re-audited after the schema-contract hardening commit. CI is green, CodeRabbit is clean, all review threads are resolved, and the command is explicit that supported selectors are Iris resource id or bundle-style productId; public numeric ASC IAP id remains intentionally unsupported until a proper REST-to-Iris bridge exists.

rudrankriyam (approved):
Re-audited latest head after the IAP schema, idempotency, and id-vs-productId collision fixes. CI is green, CodeRabbit is clean, and all review threads are resolved.

rudrankriyam:
Thanks a ton for debugging this and for the concrete schema details. That made the actual Iris/public-ID mismatch much clearer, and the follow-up is in a much better shape because of it.

PR #1661, opened 2026-06-16
Fix next-build-number with zero build upload

osrufung:
## Summary

Fix `asc builds next-build-number` so completed or expired build uploads with `cfBundleVersion` set to `0` do not make the command fail while calculating the next build number.

## Why This Is Needed

App Store Connect can leave behind completed/expired build upload records whose `cfBundleVersion` is `0`. We saw this with a real upload returned by the build upload API. Today, `next-build-number --exclude-expired` still scans build uploads after filtering processed builds, parses every upload build number, and fails with:

```text
failed to paginate build uploads: page 1: build upload <id> build number "0" must be >= 1
```

That means a stale invalid upload can break build-number automation even though `0` can never be a valid next-build-number candidate. The command should ignore that upload and continue choosing from valid processed builds, valid uploads, or the configured initial build number.

## What Changed

- Skip build upload candidates whose numeric first component is below `1` before parsing them as valid build numbers.
- Keep existing strict errors for malformed non-numeric build upload values, so unexpected bad API data still surfaces.
- Add a CLI-level regression test covering `next-build-number --exclude-expired` with an upload whose `cfBundleVersion` is `0`.

## Behavior

Example invocation that previously failed:

```bash
asc builds next-build-number --app <appid> --version <version> --exclude-expired
```

With this change, an upload with `cfBundleVersion: "0"` is ignored as an invalid candidate instead of aborting the command.

## Validation

- `PATH="$(go env GOPATH)/bin:$PATH" make format`
- `make check-command-docs`
- `make lint` (used repo fallback to `go vet ./...` because `golangci-lint` is not installed locally)
- `ASC_BYPASS_KEYCHAIN=1 go test ./internal/cli/cmdtest -run 'TestBuildsNextBuildNumber' -count=1`
- `ASC_BYPASS_KEYCHAIN=1 go test ./internal/cli/shared -run 'Test.*BuildNumber|Test.*BuildTimestamp' -count=1`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved build number calculation by skipping app store build uploads with non-positive `CFBundleVersion` values (e.g., `"0"`), preventing them from interfering with determining the next valid build number.
  * Added a CLI test covering scenarios where only non-positive build upload versions are returned, ensuring the next build number is computed correctly.

### DIFF
--- a/internal/cli/cmdtest/builds_next_number_test.go
+++ b/internal/cli/cmdtest/builds_next_number_test.go
@@ -286,7 +286,10 @@ func TestBuildsNextBuildNumberSkipsNonPositiveBuildUploadNumber(t *testing.T) {
 				t.Fatalf("expected filter[cfBundleShortVersionString]=1.2.3, got %q", query.Get("filter[cfBundleShortVersionString]"))
 			}
 			return jsonHTTPResponse(http.StatusOK, `{
-				"data":[{"type":"buildUploads","id":"expired-upload","attributes":{"cfBundleVersion":"0"}}],
+				"data":[
+					{"type":"buildUploads","id":"expired-upload","attributes":{"cfBundleVersion":"0"}},
+					{"type":"buildUploads","id":"spaced-zero-upload","attributes":{"cfBundleVersion":"0 .1"}}
+				],
 				"links":{"next":""}
 			}`), nil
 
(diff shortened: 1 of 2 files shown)


rudrankriyam:
Smart!

rudrankriyam (approved):
Reproduced the zero-upload failure on main, verified the fix and whitespace edge case, and ran the required local and CI checks. Ready to merge.

PR #1677, opened 2026-06-23
Add --include flag to testflight crashes/feedback list

DylanBettermannDD:
## Summary

Adds an opt-in `--include` flag to `asc testflight crashes list` and `asc testflight feedback list`.

The beta-feedback submission endpoints (`betaFeedbackCrashSubmissions`, `betaFeedbackScreenshotSubmissions`) don't carry the app version or build number in their attributes — but both expose a `build` relationship. Today the CLI never requests it, so there's no way to learn which build a crash or piece of feedback came from without separately parsing the symbolicated crash log.

This flag lets callers pull that relationship in:

```
asc testflight crashes list   --app <APP_ID> --include build
asc testflight feedback list  --app <APP_ID> --include build,tester
```

When `build` is included, the request also sets `fields[builds]=version,preReleaseVersion`, so the response carries the build number (`CFBundleVersion`) directly and a `preReleaseVersion` pointer the caller can resolve to the marketing version (`CFBundleShortVersionString`). This mirrors how `asc builds list` already surfaces the marketing version via `include=preReleaseVersion`.

Allowed values are `build` and `tester` — the only relationships the App Store Connect API accepts in `include` for these endpoints (verified against the embedded OpenAPI schema).

## Why

`betaFeedbackCrashSubmissions`/`betaFeedbackScreenshotSubmissions` attributes contain device/OS/locale metadata but no version/build, while the API models the link via the `build` relationship (the endpoints even support `filter[build]` / `filter[build.preReleaseVersion]`). Exposing `include` closes that gap so the build a submission was recorded against can be retrieved structurally rather than scraped from log text.

## Notes

- **Sparse-fieldset interaction:** when `--include-screenshots` is combined with `--include`, the requested relationship names are added to `fields[betaFeedbackScreenshotSubmissions]`. Without this, the sparse screenshot fieldset would omit `build`/`tester` and the API would drop the included resources from the response. Covered by a regression test.
- **Backward compatible:** with no `--include`, neither `include` nor `fields[builds]` is sent, so existing output is unchanged.
- **Pagination:** `--paginate` aggregates the `included` resources across pages, and `--next` reuses the server's `links.next` URL (which already encodes `include`/`fields`).

## Tests

- Query-builder unit tests for `include=build`, `include=tester`, combined `build,tester`, and the no-include default (asserting `fields[builds]` is set only when `build` is included).
- Regression test for `--include-screenshots` + `--include build` keeping the relationship in the screenshot fieldset.
- `ValidateInclude` unit tests (empty / single / multiple / whitespace / unknown / mixed).

All existing tests pass (`go build ./...`, `go vet`, gofumpt, full suite including `cmdtest`, and the command-docs sync check).

## Summary by CodeRabbit

* **New Features**
  * Added a new `--include` flag to both the crashes and feedback list commands (supported values: `build`, `tester`).
  * When specified, the output fetches and includes the requested related data.

* **Bug Fixes**
  * Improved query-building so requested relationships are preserved when combined (including screenshot + build scenarios).
  * Ensures build-related fields are returned only when `build` is included.

* **Tests**
  * Added unit and CLI tests for include parsing, correct request parameters, and invalid `--include` handling.

### DIFF
--- a/internal/asc/client_options.go
+++ b/internal/asc/client_options.go
@@ -403,6 +403,18 @@ func WithFeedbackIncludeScreenshots() FeedbackOption {
 	}
 }
 
+// WithFeedbackInclude specifies related resources to include in screenshot
+// feedback responses (e.g., "build" to surface the build the feedback was
+// submitted against, or "tester").
+func WithFeedbackInclude(include []string) FeedbackOption {
+	return func(q *feedbackQuery) {
+		normalized := normalizeList(include)
+		if len(normalized) > 0 {
+			q.include = normalized
+		}
+	}
+}
+
 // WithCrashDeviceModels filters crashes by device model(s).
 func WithCrashDeviceModels(models []string) CrashOption {
 	return func(q *crashQuery) {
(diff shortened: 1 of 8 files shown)


rudrankriyam (approved):
Audited the full diff and resolved threads; local full gates, repeated race coverage, built-binary validation, and credential-backed live include queries are clean.

PR #1680, opened 2026-06-24
fix(webhooks): avoid shutdown channel race

rudrankriyam:
## Summary

- capture the webhook event queue before launching workers
- add a deterministic regression for immediate worker shutdown

## Why

Workers previously evaluated `r.eventQueue` after their goroutine started, while shutdown closed that channel and replaced the field with `nil`. That caused a data race and could leave late-starting workers blocked forever on a nil channel.

Capturing the queue once preserves the existing `nil` sentinel used to reject post-shutdown enqueue attempts without making workers read mutable shutdown state. Leaving the closed channel in the field would risk send-on-closed-channel panics; locking each worker read would add lifetime coupling without helping event processing.

## Behavior

The CLI interface is unchanged. `asc webhooks serve --port 0 --output json` still reports its listener and exits successfully on interrupt; shutdown no longer races worker startup.

## Verification

- RED: existing webhook serve tests under `go test -race` reported the `startWorkers` / `stopWorkers` field race
- RED: the new immediate-stop regression timed out with workers blocked on a nil channel
- GREEN: focused regression x20 and focused race suite x10
- `ASC_BYPASS_KEYCHAIN=1 go test -race ./internal/cli/webhooks`
- Windows compile-only test for `./internal/cli/webhooks`
- `make format`
- `make check-command-docs`
- `make lint`
- pre-commit `ASC_BYPASS_KEYCHAIN=1 go test -short ./...`
- built `/tmp/asc-webhook-race`; help exited 0; live serve + interrupt exited 0

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved webhook processing to handle events with multiple workers, which can increase throughput and responsiveness.
  * Made worker shutdown more reliable when stopping the service immediately after startup.
* **Tests**
  * Added coverage for rapid start-and-stop behavior to help prevent shutdown regressions.

### DIFF
--- a/internal/cli/webhooks/webhooks_serve.go
+++ b/internal/cli/webhooks/webhooks_serve.go
@@ -264,7 +264,8 @@ func (r *webhookServeRuntime) newHandler() http.Handler {
 }
 
 func (r *webhookServeRuntime) startWorkers(_ context.Context) {
-	if r.eventQueue == nil {
+	eventQueue := r.eventQueue
+	if eventQueue == nil {
 		return
 	}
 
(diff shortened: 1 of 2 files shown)


rudrankriyam:
@coderabbitai review

PR #1691, opened 2026-06-28
Fix resolved subscription price listing

andresdefi:
## Summary
- Treat undated subscription price records as current base prices in `subscriptions pricing prices list --resolved`
- Keep future scheduled prices excluded and preserve existing latest-active selection rules
- Add `--territory` to `subscriptions pricing prices list` so resolved output can be filtered to one territory

## Root cause
The resolved subscription price path fetched `GET /v1/subscriptions/{id}/prices?include=subscriptionPricePoint,territory&fields[subscriptionPricePoints]=customerPrice,proceeds,proceedsYear2&fields[territories]=currency&limit=200`, but then skipped every row whose `attributes.startDate` was empty unless a plan-type filter was set. App Store Connect can return current/base per-territory subscription prices without `startDate`; the summary command already accepts those as current. When all current territory prices are undated, the CLI reduced a non-empty API response to `{"prices":[]}`.

Request diff from the reported reproduction:
- Raw list: `GET /v1/subscriptions/<SUB_ID>/prices` returned `subscriptionPrices` data.
- Resolved list: `GET /v1/subscriptions/<SUB_ID>/prices?include=subscriptionPricePoint,territory&fields[subscriptionPricePoints]=customerPrice,proceeds,proceedsYear2&fields[territories]=currency&limit=200` returned data plus included price points/territories, but the CLI dropped undated rows during response mapping.
- With this fix, `--resolved --territory USA` adds `filter[territory]=USA` to the same resolved request shape.

## Tests
- `go test ./internal/cli/cmdtest -run 'TestSubscriptionsPricingPricesListResolved|TestSubscriptionsPricingPricesListRawOutputUnchanged'`
- `go test ./internal/cli/subscriptions`
- `go test ./internal/asc/...`
- `env -u ASC_KEY_ID -u ASC_ISSUER_ID -u ASC_KEY_PATH -u ASC_PRIVATE_KEY -u ASC_PROFILE go test ./internal/cli/shared`

Note: `go test ./internal/cli/...` passed all affected packages but failed in `internal/cli/shared` when local ASC credential/keychain state leaked into `TestResolveAuthCredentialsMetadata_FallsBackToEnvKeyID`; rerunning that package with ASC credential env cleared passed.

## Summary by CodeRabbit

* **New Features**
  * Added `--territory` filtering to `asc subscriptions prices list`, including resolved-price views, with pagination-aware filtering.
  * Resolved pricing output now includes **Plan Type** (new table column).

* **Bug Fixes**
  * Resolved pricing now returns undated “current” prices per plan type when a plan type is provided.
  * Resolved pricing honors the requested territory filter during retrieval and resolution.

* **Tests**
  * Expanded CLI and resolved-pricing tests for territory handling, plan-type behavior, updated resolution consumption, and output changes.

### DIFF
--- a/internal/cli/cmdtest/subscriptions_prices_plan_type_test.go
+++ b/internal/cli/cmdtest/subscriptions_prices_plan_type_test.go
@@ -231,6 +231,15 @@ func TestSubscriptionsPricingPricesListPlanTypeValidationErrors(t *testing.T) {
 			},
 			wantErr: "invalid value for --plan-type: cannot be empty",
 		},
+		{
+			name: "empty territory",
+			args: []string{
+				"subscriptions", "pricing", "prices", "list",
+				"--subscription-id", "8000000001",
+				"--territory", "",
+			},
+			wantErr: "invalid value for --territory: cannot be empty",
+		},
 	}
 
 	for _, test := range tests {
@@ -285,4 +294,12 @@ func TestSubscriptionsPricingPricesListPlanTypeUsageExitCodes(t *testing.T) {
 			}, test.wantErr)
 		})
 	}
+}
+
+func TestSubscriptionsPricingPricesListTerritoryUsageExitCodes(t *testing.T) {
+	assertUsageExit(t, []string{
+		"subscriptions", "pricing", "prices", "list",
+		"--subscription-id", "8000000001",
+		"--territory", "",
+	}, "invalid value for --territory: cannot be empty")
 }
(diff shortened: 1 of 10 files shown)


rudrankriyam (dismissed):
Verified the resolved subscription pricing fix and the follow-up territory validation change. Local focused tests, full test suite, docs, lint, build, and GitHub CI are green.

rudrankriyam (approved):
Verified the resolved subscription pricing fix, territory validation, and plan-aware resolved output on the latest head. Local focused tests, full validation, build, and GitHub CI are green.

PR #1710, opened 2026-07-06
[codex] Add web app delete command

rudrankriyam:
## Summary

Adds a guarded `asc web apps delete` command for removing throwaway App Store Connect apps through the web-session API.

## What changed

- Added web client support for `GET /apps/{id}` and the observed web-session delete shape: `PATCH /apps/{id}` with `attributes.removed = true`.
- Added `asc web apps delete --app APP_ID_OR_BUNDLE_ID --confirm` under the existing web apps command group.
- Added exact identity guard flags, `--expected-bundle-id` and `--expected-name`, so destructive calls can stop before mutation when the resolved app is not the intended target.
- Added CLI and HTTP coverage for confirm gating, exact bundle-ID lookup, identity mismatch behavior, command registration, and the outgoing PATCH payload.

## Behavior

Examples:

```bash
asc web apps delete --app "1234567890" --confirm
asc web apps delete --app "com.example.throwaway" --confirm
asc web apps delete --app "1234567890" --expected-bundle-id "com.example.throwaway" --confirm
```

The command does not use ambient `ASC_APP_ID` fallback. Numeric `--app` values are treated as explicit App Store Connect app IDs; non-numeric values resolve via exact bundle ID lookup.

## Validation

- `git diff --check`
- `ASC_BYPASS_KEYCHAIN=1 go test ./internal/web ./internal/cli/web ./internal/cli/cmdtest`
- `make check-docs`
- pre-commit hook on each commit: command docs, formatting, lint, and short test suite
- Live web-session smoke: command returned `removed: true` for disposable throwaway apps and follow-up `asc apps list --paginate` no longer returned those app IDs

Note: lint passed with `0 issues`; golangci-lint printed a stale cache warning referencing an unrelated deleted local worktree path.

## Summary by CodeRabbit

* **New Features**
  * Added `web apps delete` to the CLI for removing apps.
  * Supports lookup by app ID or bundle ID, includes guardrails (`--expected-bundle-id`, `--expected-name`) and requires `--confirm`.
  * Outputs structured results (including `removed:true`) and renders in table/markdown formats.

* **Bug Fixes**
  * Deletion now halts before making API calls when lookups/guards don’t match or when `--confirm` is missing.

* **Tests**
  * Added unit tests for command registration, deletion behavior, guard/confirmation enforcement, and correct handling when delete responses omit attributes.

### DIFF
--- a/internal/cli/cmdtest/web_commands_test.go
+++ b/internal/cli/cmdtest/web_commands_test.go
@@ -49,6 +49,13 @@ func TestWebAppsCreateSubcommandIsRegistered(t *testing.T) {
 	}
 }
 
+func TestWebAppsDeleteSubcommandIsRegistered(t *testing.T) {
+	root := RootCommand("1.2.3")
+	if sub := findSubcommand(root, "web", "apps", "delete"); sub == nil {
+		t.Fatalf("expected web apps delete to be registered")
+	}
+}
+
 func TestWebAppsMedicalDeviceSetSubcommandIsRegistered(t *testing.T) {
 	root := RootCommand("1.2.3")
 	if sub := findSubcommand(root, "web", "apps", "medical-device", "set"); sub == nil {
--- a/internal/cli/web/web_apps.go
+++ b/internal/cli/web/web_apps.go
@@ -32,6 +32,7 @@ Use ` + "`asc web apps create`" + ` as the canonical app-creation command.
 		UsageFunc: shared.DefaultUsageFunc,
 		Subcommands: []*ffcli.Command{
 			WebAppsCreateCommand(),
+			WebAppsDeleteCommand(),
 			WebAppsAvailabilityCommand(),
 			WebAppsCompatibilityCommand(),
 			WebAppsMedicalDeviceCommand(),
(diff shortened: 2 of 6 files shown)


rudrankriyam:
Client slice: this adds `GetApp` plus `DeleteApp` for the web-session app resource. The delete call uses the sanitized endpoint shape captured from App Store Connect: `PATCH /iris/v1/apps/{id}` with `attributes.removed = true`, and keeps the existing web client request/error handling path.

rudrankriyam:
Command slice: `asc web apps delete` requires `--confirm` before resolving a session, accepts an explicit app ID or exact bundle ID, avoids ambient `ASC_APP_ID` fallback, and supports `--expected-name` / `--expected-bundle-id` guardrails before mutation.

rudrankriyam:
Coverage slice: tests assert the PATCH payload, command registration, confirm gating before session lookup, bundle-ID lookup mismatch handling, and expected identity mismatch behavior so destructive calls fail closed.

rudrankriyam:
@coderabbitai review

rudrankriyam:
Addressed CodeRabbit's failure-path coverage note in `b05a2776`: the delete failure tests now capture stdout/stderr explicitly. Guard mismatch paths assert no output while preserving returned-error assertions, and the missing `--confirm` path asserts the expected stderr message before session resolution.

rudrankriyam:
Addressed Cursor Bugbot's stale `removed` output finding in `273bf9d0`: fallback app data is now used only for display fields, while the `removed` flag is taken from the delete response when present and otherwise stays `true` after a successful delete call. Added a regression test for sparse delete responses with pre-delete fallback data.

PR #1756, opened 2026-07-15
fix(errors): surface HTTP status in exit codes for retried and auth failures

rudrankriyam:
## Motivation

The 2.8.2 exit-code audit reproduced two bugs where the CLI exits with the wrong code despite `concepts/error-handling.mdx` documenting deterministic HTTP-status mapping:

1. **HIGH — retried statuses lose their HTTP status.** `buildRetryableError` (internal/asc/client_http.go) constructed the retry error with `errors.New(message)`, discarding the parsed `*APIError`. After retries exhaust (GET/HEAD) or immediately (mutating methods), the final `*asc.RetryableError` exposed no HTTP status: a persistent 503 exited with generic code 1 instead of the documented `ExitHTTPServiceUnavailable` (63), and `httpStatusFromError` returned 0, leaving telemetry blind to the entire 5xx/429 class.
2. **MEDIUM — Apple 401s don't map to ExitAuth.** `APIError.Is` matched `ErrUnauthorized`/`ErrForbidden` only by the exact code strings `UNAUTHORIZED`/`FORBIDDEN`, but Apple actually returns 401 with code `NOT_AUTHORIZED`. Real auth failures therefore exited 11 instead of the documented `ExitAuth` (3).

## Approach

- `buildRetryableError` now wraps the parsed `*APIError` (or a synthesized one carrying the status when the response body is empty, e.g. the upload retry path) behind the existing human-readable message via a small `retryableStatusError` with `Unwrap()`. The rendered message is unchanged; `errors.Is`/`errors.As`, `ExitCodeFromError`, and telemetry now see the status through the chain, including through the `retry limit exceeded ... %w` wrapper.
- `RetryableError` gained `HTTPStatusCode()` delegating to the wrapped chain (0 for transport failures).
- `APIError.Is` now also matches `ErrUnauthorized` on `StatusCode == 401` and `ErrForbidden` on `StatusCode == 403`, mirroring the existing 404 handling; code-string matching is kept.
- Updated the 401/403 rows in `concepts/error-handling.mdx` to document the ExitAuth special case, matching the 404/409 rows.

## Behavior change (targeted at 3.0.0)

This intentionally changes exit codes:

- Persistent retryable statuses now exit with their documented mapped codes instead of 1 (503→63, 500→60, 502→62, 504→64, 429→39, 408→18), and telemetry now records their `http_status`.
- 401/403 responses now exit 3 (`ExitAuth`) instead of 11/12, regardless of the `code` string Apple sends.

Scripts matching on exit code 1 or 11/12 for these failures will observe the new values.

## Tests

- `internal/asc/errors_test.go`: `APIError.Is` with the real Apple payload shape (`{"errors":[{"code":"NOT_AUTHORIZED",...}]}`, status 401), a 403 payload, and code-string-only matching without a status.
- `internal/asc/client_http_test.go`: retry-exhaustion through `GetApps` asserting the final error still satisfies `IsRetryable`, exposes `*APIError` with status 503, and reports `HTTPStatusCode()`; `buildRetryableError` with a nil body (upload path) keeps the exact previous message while carrying the status.
- `cmd/exit_codes_test.go`: end-to-end through `httptest` servers (existing tests constructed `APIError` directly, bypassing the retry wrapper): persistent 503 → exit 63, persistent 429 → exit 39, and a NOT_AUTHORIZED 401 → exit 3, each also asserting `httpStatusFromError` for telemetry.

Verified with `go build ./...`, `go vet` on touched packages, `make format-check`, `make check-docs`, `make lint`, and `ASC_BYPASS_KEYCHAIN=1 go test ./internal/asc ./cmd ./internal/cli/...` (all green). Full-suite runs on the shared dev machine intermittently flake on unrelated timing-sensitive blackbox tests (`TestRun_BuiltBinaryEmitsSchemaV4Payload`, `TestWinBackOffersLinksFallsBackToNumericIDAfterLookupTimeout`); both pass in isolation and in full package runs.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved retry-exhausted HTTP errors so the originating HTTP status and parsed API error are retained for accurate exit-code and telemetry handling.
  - Updated authentication detection to consistently treat Apple `401 NOT_AUTHORIZED` as an authentication failure.
  - Adjusted 4xx exit-code mapping: both `401 Unauthorized` and `403 Forbidden` now map to the authentication-failure exit code.

- **Documentation**
  - Refreshed the HTTP status-to-exit-code guidance to reflect the updated authentication mappings.

- **Tests**
  - Added coverage for retry-exhausted error behavior and authentication matching via HTTP status and API error codes.


rudrankriyam:
@coderabbitai review

PR #1765, opened 2026-07-15
fix(web): classify invalid credentials as auth errors

rudrankriyam:
## Summary

- map wrapped `webcore.ErrInvalidAppleAccountCredentials` errors to the documented auth exit code 3
- preserve the existing stderr message and web-login behavior
- add a regression at the global exit-code boundary

## Why

ASC CLI v4 telemetry already classifies rejected Apple Account credentials as `auth_error`, but the process returned generic exit code 1. The web login path preserves a dedicated sentinel, so this can be classified narrowly without changing transport, server, or other web errors.

This PR is stacked on #1763 because that change introduces the shared `webcore` dependency in the exit mapper.

## Verification

- RED: the wrapped credentials sentinel returned exit 1 before the mapper change
- GREEN: focused `TestExitCodeFromError` regression returns exit 3
- `make format`
- `make check-docs`
- `make lint`

## Summary by CodeRabbit

* **Bug Fixes**
  * Invalid Apple account credentials are now consistently reported as authentication failures.
  * Wrapped credential errors now map to the correct authentication exit status.


rudrankriyam:
@coderabbitai review

PR #1783, opened 2026-07-17
fix(subscriptions): require adjusted equalization filters

rudrankriyam:
## What changed

- Require `--upfront-price-point-id` and `--plan-type MONTHLY` for first-page adjusted-equalizations requests.
- Return exit 2 before client construction when either live-required filter is absent.
- Keep opaque `--next` URLs self-contained; owner and query flags remain forbidden with `--next`.
- Update command help, examples, and regression coverage.

## Why

Apple documents both values as optional query filters in the 4.4.1 OpenAPI snapshot, but the live endpoint rejects requests without the pair. A request with neither filter reports a missing required filter; `--plan-type MONTHLY` without an upfront point reports an invalid upfront point. The same request succeeds when both values are present.

The CLI now catches that server requirement before auth or HTTP instead of sending a request that cannot work.

## Verification

RED showed both missing-filter cases reaching the client factory and exiting 1. The new tests make each case exit 2 before client construction, cover the exact query, retain opaque-next pagination, and reject next/filter conflicts.

The review follow-up runs the sparse-field path and query assertions through a TLS `httptest.Server`. Its test transport keeps certificate verification enabled while dialing the local server.

Ran on exact base `b92e560549ae9d314164e686205af1fdf2bea3dd`:

- `make format`
- `make check-docs`
- `make lint`

A read-only call from the branch binary against the throwaway-account price point succeeded with both required filters and `--limit 1`, returning one item from 174 adjusted equalizations.

## Summary by CodeRabbit

* **New Features**
  * Enhanced `adjusted-equalizations` command help/usage to clearly mark `--upfront-price-point-id` and `--plan-type MONTHLY` as required, including explicit guidance for first-page requests.
  * Added clearer messaging that required owner flags can’t be combined with `--next`.

* **Bug Fixes**
  * Added upfront validation to error when required live filter flags are missing, preventing invalid requests before contacting the service.
  * Updated adjusted equalizations request filtering to include the corresponding live filters.

* **Tests**
  * Expanded and refactored CLI tests to verify help text, required-filter enforcement, and correct filter inclusion.

### DIFF
--- a/internal/cli/cmdtest/subscriptions_adjusted_equalizations_test.go
+++ b/internal/cli/cmdtest/subscriptions_adjusted_equalizations_test.go
@@ -5,9 +5,11 @@ import (
 	"encoding/json"
 	"errors"
 	"io"
+	"net"
 	"net/http"
 	"net/http/httptest"
 	"net/url"
+	"os"
 	"testing"
 
 	"github.com/rudrankriyam/App-Store-Connect-CLI/internal/asc"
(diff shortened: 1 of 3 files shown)


rudrankriyam:
@coderabbitai review

PR #1836, opened 2026-07-31
feat(analytics): add report instance filters

ayushhagarwal:
## Summary

- expose `WithAnalyticsReportInstancesGranularity` for the documented
  `filter[granularity]` query parameter
- expose `WithAnalyticsReportInstancesProcessingDates` for the documented
  `filter[processingDate]` query parameter
- normalize whitespace, ignore empty dates, remove duplicates, and preserve
  caller order
- add a focused request-level test covering the encoded filters, pagination
  limit, authentication header, and response decoding

## Why

`GetAnalyticsReportInstances` currently exposes pagination but not Apple's
instance filters. Callers that need a particular weekly or daily report must
therefore fetch and scan every available instance or rebuild request URLs
outside the client.

This adds the missing, reusable client primitive without changing existing
behavior. It also keeps filtering on Apple's API rather than duplicating it in
each downstream workflow.

## Scope

This PR is intentionally narrow. It does not add CLI flags, create analytics
report requests, download report segments, aggregate metrics, or change output
formats.

## End goal: evidence-backed investor presentations

The intended end state is an optional `asc-app-dossier` skill in the
[companion skills repository](https://github.com/rorkai/app-store-connect-cli-skills).
An app owner could ask it to prepare an acquisition, fundraising, or
investor-update presentation from their App Store evidence.

```text
Apple Analytics Reports
  -> deterministic asc metrics with provenance and privacy caveats
  -> optional asc-app-dossier skill
  -> evidence ledger + written memo + presentation-ready slide deck
```

The presentation would separate every material claim into `verified`,
`derived`, `owner-provided`, `assumption`, or `unavailable`. It would not invent
traction, conversion, retention, revenue, profit, market size, valuation, or
transfer readiness when the evidence does not support those claims.

The core CLI would remain generic and deterministic; presentation generation
would stay in the optional skill rather than becoming CLI behavior.

## Proposed cross-repository sequence

1. **This PR:** expose the documented filters needed to select exact weekly
   report instances.
2. **Separate CLI PR:** download, integrity-check, and parse all compressed
   segments, then emit additive weekly metrics with provenance and explicit
   `unavailable` states.
3. **Separate skills PR:** add `asc-app-dossier`, consuming that stable JSON to
   create an evidence ledger, investor/acquisition memo, chart plan, and
   presentation-ready slide deck.

I would appreciate maintainer feedback on this boundary and sequence before
submitting the larger follow-ups. Each step remains independently useful and
reviewable.

## Validation

Validated locally with the repository's required Go 1.26 toolchain:

- `go test ./internal/asc`
- focused race tests
- `make format`
- `make lint`
- `make check-docs`
- `git diff --check main...HEAD`

### DIFF
--- a/internal/asc/analytics.go
+++ b/internal/asc/analytics.go
@@ -210,6 +210,8 @@ type analyticsReportsQuery struct {
 
 type analyticsReportInstancesQuery struct {
 	listQuery
+	granularity    string
+	processingDate []string
 }
 
 type analyticsReportSegmentsQuery struct {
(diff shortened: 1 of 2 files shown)


rudrankriyam:
Thanks for laying out the sequence. I understand the intended workflow now.

I am still closing this draft because step 1 is not independently useful in this repository. The new helpers live under `internal/asc`, no command calls them today, and the skills repository cannot import a Go `internal` package. The first user-visible CLI change is step 2.

Please bring the filter work back with the CLI PR that exposes the actual command path. One concrete version would update `asc analytics view --request-id "REQUEST_ID" --date "2024-01-20" --paginate` so it sends `filter[processingDate]` to Apple instead of fetching every instance and filtering locally. Command-level tests should cover the request, pagination, and unchanged output. If weekly granularity is also needed, expose it through that command and model the documented multi-value `DAILY`, `WEEKLY`, and `MONTHLY` contract.

The dossier skill can consume the stable JSON after that CLI contract exists. Keeping the helper and its first caller in one PR will let us review the behavior end to end.

ayushhagarwal:
Thank you @rudrankriyam for a detailed feedback on the idea. I agree that helper only PR was not independently useful. I’ll return with one focused PR that wires `processingDate` filtering into analytics view, models the documented multi-value contract, preserves existing output, and _adds command-level request and pagination coverage_. 
I’ll keep the dossier skill separate until that CLI contract is established.
Really appreciate the chance to contribute to this.

PR #1864, opened 2026-08-05
fix(telemetry): drop permanently rejected events

rudrankriyam:
## Summary

Treat collector rejections that replay cannot repair as permanent. Telemetry events rejected with a 4xx response are acknowledged and removed from the bounded local spool, except HTTP 408 and HTTP 429, which remain retryable alongside network errors and 5xx responses.

This prevents an unsupported schema version or semantic event shape from poisoning later telemetry flushes while preserving retries for transient failures. The collector contract documentation now states the same boundary.

## Rollout relationship

Companion Rork collector hardening: https://github.com/rorkai/rorkai/pull/4914

The two PRs can merge independently, but this CLI behavior should land before any future schema or semantic event expansion. For a future root command, apply the Rork constraint migration, deploy the Rork API normalization, and only then release that CLI command.

Neither PR releases or deploys anything.

## Testing

- `make build`
- `make format`
- `make check-docs`
- `make lint`
- Focused client and worker tests cover permanent 400/422 responses, retryable 408/429/500 responses, and spool removal without retry poison.

### DIFF
--- a/commands/telemetry.mdx
+++ b/commands/telemetry.mdx
@@ -164,8 +164,10 @@ Collectors should accept `POST /asc/v1/events`, validate the event schema, add
 server-side receive metadata, and insert accepted events into analytics storage.
 They should return a 2xx response only after that insert commits; temporary
 storage failures should return a retryable non-2xx response so the CLI keeps the
-event in its bounded local spool. Collector credentials must never be shipped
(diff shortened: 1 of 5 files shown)


rudrankriyam:
@coderabbitai review

PR #1992, opened 2026-08-13
feat: add direct apps rename command

rudrankriyam:
## Summary

- add `asc apps rename --app APP_ID --locale LOCALE --name NAME`
- update an existing app-info localization or create the requested locale when it is missing
- share the localization target planner with `app-setup info set`, preserving the existing command
- support JSON, table, and Markdown results with the resolved App Info and localization IDs

## Behavior

```bash
asc apps rename --app "APP_ID" --locale "en-US" --name "New Name"
```

`--app-info` remains available as an explicit override when an app has multiple App Info records. Required flags, locale syntax, and the 30-character Unicode name limit are checked before client creation. The planner rejects duplicate locale matches and empty localization IDs before any write.

The API path matches the OpenAPI snapshot:

- `GET /v1/appInfos/{id}/appInfoLocalizations?filter[locale]=...`
- `PATCH /v1/appInfoLocalizations/{id}` with only `name`
- `POST /v1/appInfoLocalizations` with required `locale`, `name`, and App Info relationship

## Validation

- `make format`
- `make check-docs`
- `make lint`
- built-binary help and pre-auth exit-code checks
- command-level HTTP tests for update, create, API failure, explicit App Info selection, duplicate and blank targets, flag ordering, JSON, table, and Markdown output

During a missing-flag black-box check, removing the app environment variable allowed the local config default to resolve. Apple rejected the requested name as unavailable, and a read-only follow-up confirmed that no localization changed. Subsequent missing-app verification used an explicitly empty environment value and returned exit code 2 before client creation.

## Summary by CodeRabbit

* **New Features**
  * Added the experimental `apps rename` command to create or update localized app names.
  * Supports app, locale, name, and optional App Info selection with validation.
  * Provides JSON, table, and Markdown output, including whether a localization was created or updated.

* **Bug Fixes**
  * Improved handling of missing, empty, duplicate, or mismatched localizations.
  * Prevents updates to App Info belonging to another app.

* **Documentation**
  * Added command usage, options, examples, and response field documentation.

### DIFF
--- a/commands/apps.mdx
+++ b/commands/apps.mdx
@@ -177,8 +177,9 @@ does not exist, the command creates its app-info localization.
   App Info ID override. Use this when the app has multiple App Info records.
 </ParamField>
 
-<ParamField path="--output" type="string" default="json">
-  Output format: `json`, `table`, `markdown`
+<ParamField path="--output" type="string">
+  Output format: `json`, `table`, `markdown`. Defaults to `table` in an
+  interactive terminal and minified `json` for pipes or CI.
 </ParamField>
 
 **Examples:**
--- a/commands/apps.mdx
+++ b/commands/apps.mdx
@@ -156,10 +156,11 @@ asc apps view --id "APP_ID" --output table
 
 ***
 
-### `asc apps rename`
+### `asc apps rename` (experimental)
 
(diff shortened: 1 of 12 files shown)


rudrankriyam:
@coderabbitai review

PR #2045, opened 2026-08-16
fix(submit): include the appStoreVersion relationship instead of only naming it in fields[]

Radwan-Albahrani:
## Problem

#2027 did not fix `asc review submit`. Running **4.4.2** (which contains that fix) still aborts on its final call:

```
Error: review submit: submit review: final submission validation:
review submission 3f07316f-1770-4675-96db-07b0eb857f55 does not contain target version 5434872c-4d84-4d72-af52-9da1a46d62ee
```

## Why the previous fix was wrong

#2027 correctly identified the function and the missing linkage, but reached for the wrong query parameter:

```go
asc.WithReviewSubmissionItemsFields([]string{"appStoreVersion"})
```

`fields[...]` is a **sparse-fieldset selector**. It narrows which fields come back; it does not ask the API to materialise relationship *linkage*. Measured against the live API on the submission above:

`?fields[reviewSubmissionItems]=appStoreVersion&limit=200`

```json
{ "data": [ { "type": "reviewSubmissionItems",
              "id": "M2YwNzMxNmYt…",
              "links": { "self": "…" } } ],
  "meta": { "paging": { "total": 1, "limit": 200 } } }
```

No `relationships` key at all. Note `total: 1` — the item was always present; only its linkage was missing. So `reviewSubmissionItemVersionID` still read `""`, `hasTargetVersion` stayed false, and the guard rejected a correctly prepared submission exactly as before.

`?include=appStoreVersion&limit=200` on the same submission:

```json
"relationships": {
  "appStoreVersion": { "data": { "type": "appStoreVersions",
                                 "id": "5434872c-4d84-4d72-af52-9da1a46d62ee" } } }
```

That is the target version id. `include=` is what makes App Store Connect emit the linkage.

## Why #2027's test stayed green

Its stub returns `relationships` unconditionally:

```go
return submitJSONResponse(http.StatusOK, `{
    "data": [{ "type": "reviewSubmissionItems", "id": "item-1",
               "relationships": { "appStoreVersion": { "data": {…} } } }]
}`)
```

A stub that always supplies linkage cannot tell a request that asks for it from one that does not, so it pins the *request shape* and never the API's actual reply. The PR reasoned that the real API returns relationships "on request" — true, but only for `include=`, and the test could not check that.

## The fix

Send `include=appStoreVersion` at the three call sites that read `appStoreVersion.data.id`; `fields[]` rides along to keep the payload narrow.

- `internal/cli/submit/submit_create.go` — `summarizeReviewSubmissionItems`, the one that breaks `review submit`
- `internal/cli/submit/submit_status.go` — `resolveReviewSubmissionVersionIDFromItems`, same bug, fails soft by reporting "no version"
- `internal/cli/reviews/review_overview.go` — same bug, feeds `reviewItemMatchesVersion`

`internal/cli/backgroundassets/background_assets_submit.go` and `internal/cli/reviews/{submissions_history,review_items}.go` already pair `Fields` with `Include`, which is the pattern this brings the rest in line with.

## Verification

**1. Against the live API.** A temporary GET-only probe (transport rejected any non-GET) called the real `summarizeReviewSubmissionItems` against submission `3f07316f…`:

| build | request | result |
|---|---|---|
| 4.4.2 as shipped | `fields[]` only | `hasTargetVersion=false hasOtherItems=true` — the production bug |
| this PR | `fields[]` + `include=` | `hasTargetVersion=true hasOtherItems=false` |

Same code, same submission; the `include` is the only difference. The probe is not part of this PR.

**2. Regression test that actually fails first.** `submit_create_items_include_test.go` stubs the API's *observed* behaviour — linkage only when `include=appStoreVersion` is present, otherwise the captured link-only payload. Against `main` it fails; with the fix it passes. A second test pins the link-only payload as unusable so nobody swaps `include` back for `fields`.

Checked before/after: `#2027`'s own test passes against unfixed code, which is precisely the blind spot this replaces.

**3.** `go build ./...`, `go vet`, and `go test ./internal/cli/submit/ ./internal/cli/reviews/` all pass. Four pinned request URLs in `submit_test.go` were updated for the new query string.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved review-submission processing to reliably identify the associated App Store version.
  * Fixed version-specific filtering and status detection when retrieving submission items.

* **Tests**
  * Added regression coverage to verify App Store version relationships are correctly retrieved and resolved.

### DIFF
--- a/internal/cli/reviews/review_overview.go
+++ b/internal/cli/reviews/review_overview.go
@@ -405,6 +405,7 @@ func summarizeReviewSubmissionItems(ctx context.Context, client *asc.Client, sub
 		ctx,
 		submissionID,
 		asc.WithReviewSubmissionItemsLimit(200),
+		asc.WithReviewSubmissionItemsInclude([]string{"appStoreVersion"}),
 		asc.WithReviewSubmissionItemsFields([]string{"state", "appStoreVersion"}),
 	)
 	if err != nil {
--- a/internal/cli/submit/submit_create.go
+++ b/internal/cli/submit/submit_create.go
@@ -292,12 +292,16 @@ func summarizeReviewSubmissionItems(
 		return summary, nil
 	}
 
-	// appStoreVersion must be requested explicitly. Without it the API returns items carrying no
-	// relationship linkage, so every item reads as an empty version id, hasTargetVersion never
-	// becomes true, and a correctly prepared submission is rejected as not containing its version.
+	// appStoreVersion must be INCLUDED, not merely named in fields[]. fields[] is a sparse-fieldset
+	// selector: it narrows what comes back, and the API answers it with items that carry "links" only
+	// and no "relationships" key at all. Every item then reads as an empty version id,
+	// hasTargetVersion never becomes true, and a correctly prepared submission is rejected as not
+	// containing its version. include= is what makes App Store Connect materialise the linkage;
+	// fields[] rides along to keep the item payload narrow.
 	resp, err := client.GetReviewSubmissionItems(
 		ctx,
 		submissionID,
+		asc.WithReviewSubmissionItemsInclude([]string{"appStoreVersion"}),
 		asc.WithReviewSubmissionItemsFields([]string{"appStoreVersion"}),
 		asc.WithReviewSubmissionItemsLimit(200),
 	)
--- a/internal/cli/submit/submit_create_items_include_test.go
+++ b/internal/cli/submit/submit_create_items_include_test.go
@@ -1,0 +1,114 @@
+package submit
+
+import (
+	"context"
+	"net/http"
+	"testing"
+)
+
+// respondLikeAppStoreConnectItems replays what the live API actually does for
+// GET /v1/reviewSubmissions/{id}/items, captured on 2026-08-16 against submission
+// 3f07316f-1770-4675-96db-07b0eb857f55:
+//
+//   - ?fields[reviewSubmissionItems]=appStoreVersion  -> the item carries "links" ONLY. There is no
+//     "relationships" key at all, because fields[] is a sparse-FIELDSET selector and does not ask
+//     the API to materialise relationship linkage.
+//   - ?include=appStoreVersion                        -> the item carries
+//     relationships.appStoreVersion.data.id.
+//
+// Stubbing that difference is the whole point. A stub that returns relationships unconditionally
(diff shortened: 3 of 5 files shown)


Radwan-Albahrani:
## End-to-end proof against the live App Store Connect account

Built this branch and ran the real `asc review submit` against a production app (Kaset, `6787275694`, version 1.3.1, build 154). This is the exact call that failed on 4.2.0 and again on 4.4.2.

**Setup** — cancelled the in-flight submission so the version was submittable again:

```
version 5434872c-4d84-4d72-af52-9da1a46d62ee -> DEVELOPER_REJECTED
reviewState COMPLETE
```

**Then, with this branch's binary:**

```
$ asc review submit --app 6787275694 --version 1.3.1 \
    --build 995255e8-a534-4a98-aeca-44a5c2cf832a --platform IOS --confirm

{"appId":"6787275694","version":"1.3.1",
 "versionId":"5434872c-4d84-4d72-af52-9da1a46d62ee",
 "buildId":"995255e8-a534-4a98-aeca-44a5c2cf832a",
 "submissionId":"894452c6-3f39-4c6f-9ae9-eece9999e45a",
 "submittedDate":"2026-08-16T16:07:16.081Z", ...}
```

No error, and the command completed its own final submission validation — the step that previously aborted with `does not contain target version`. Zero occurrences of `error` or `does not contain` in the output.

**Resulting state:**

```json
{ "version":          { "state": "WAITING_FOR_REVIEW" },
  "latestSubmission": { "id": "894452c6-3f39-4c6f-9ae9-eece9999e45a",
                        "state": "WAITING_FOR_REVIEW" },
  "reviewState": "WAITING_FOR_REVIEW",
  "nextAction": "Wait for App Store review outcome." }
```

For contrast, the two prior releases of the same app on unfixed builds both aborted at that step and had to be finished by hand with `asc review submissions-submit --id <id> --confirm`. This is the first time `asc review submit` has completed the submission on its own here — no fallback, no manual step.

Radwan-Albahrani:
@rudrankriyam i apologize for my Claude he didn't test the first one properly. 

This time hopefully fixed.

I told him to patch and rebuild the binary on my device and test it and it seems to actually run to the end this time.

rudrankriyam (approved):
Audited the exact head against current main. The API contract and live-reported behavior confirm that fields[] alone does not materialize relationship linkage; include=appStoreVersion is required. Focused and full local gates pass, and the regression covers the prior blind spot.

PR #2066, opened 2026-08-18
Reject unsafe build artifacts before upload reservation

rudrankriyam:
## What changed

- Require IPA and PKG upload inputs to be non-empty regular files.
- Reject PKG symlinks instead of following them.
- Apply the checks before authentication and build-upload reservation.

## Why

PKG uploads previously used a following `stat` call and accepted zero-byte or non-regular paths. That could dereference an unintended target or create remote upload resources for an artifact that cannot be uploaded successfully. IPA checks now use the same complete file contract for consistent behavior.

## Compatibility

Valid regular IPA and PKG files are unchanged. Empty files, symlinks, directories, and other non-regular inputs now fail locally with a flag-specific error.

## Validation

- Captured zero-byte and symlinked PKG reservations as RED CLI/HTTP tests.
- Added shared validator coverage for empty IPA/PKG files and PKG symlinks.
- Ran focused build, publish, shared, and CLI suites.
- `make format`
- `make check-docs`
- `make lint`

## Summary by CodeRabbit

* **Bug Fixes**
  * PKG and IPA uploads now reject empty files, directories, non-regular files, and symlinked paths before network requests.
  * Uploads now use the validated artifact contents throughout transfer and checksum verification, even if the artifact path changes.
  * Upload validation errors remain clearly associated with the builds upload command.

* **Tests**
  * Added coverage confirming invalid PKG and IPA artifacts are detected during validation.
  * Added coverage confirming uploads remain tied to the originally validated artifact.

### DIFF
--- a/internal/asc/upload.go
+++ b/internal/asc/upload.go
@@ -301,6 +301,20 @@ func VerifySourceFileChecksums(filePath string, expected *Checksums) (*Checksums
 	if expected == nil {
 		return nil, nil
 	}
+	file, err := openUploadSourceFile(filePath)
+	if err != nil {
+		return nil, fmt.Errorf("open file for checksum: %w", err)
+	}
+	defer file.Close()
+	return VerifySourceFileChecksumsFromFile(file, expected)
+}
+
+// VerifySourceFileChecksumsFromFile computes and compares API-provided
+// checksums against an already-opened source file.
+func VerifySourceFileChecksumsFromFile(file *os.File, expected *Checksums) (*Checksums, error) {
+	if expected == nil {
+		return nil, nil
+	}
 
 	computed := &Checksums{}
 	if expected.File != nil {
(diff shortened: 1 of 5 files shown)


rudrankriyam:
Addressed all three review findings in `b2080f51` after reproducing the path-replacement race:

- opens IPA/PKG artifacts once with no-follow semantics and verifies the opened inode matches the validated path
- keeps that handle through metadata extraction, reservation sizing, upload operations, and checksum verification
- includes `--ipa` or `--pkg` in symlink errors
- makes both new symlink fixtures fail, rather than skip, when setup fails
- adds a CLI regression that replaces the path after reservation and proves both upload bytes and committed checksum still come from the originally validated file

Validation completed locally:

- focused regression and affected-package tests
- race-enabled tests for `internal/asc`, `internal/cli/builds`, `internal/cli/shared`, and `internal/cli/cmdtest`
- `make format`
- `make check-docs`
- `make lint`

rudrankriyam:
@codex review

PR #2213, opened 2026-08-26
fix(json): preserve null sparse resource attributes

rudrankriyam:
## Summary

- decode an explicitly present `attributes: null` member into the typed resource attributes value
- preserve the generic resource's existing absent/null/output behavior
- cover custom attribute decoders that need to observe a present null value

## Why

The bundle ID query-parity and generic sparse-resource changes exposed an integration bug on current `main`. `Resource[T].UnmarshalJSON` recorded `attributes: null` but returned before invoking a custom `T.UnmarshalJSON`. The bundle ID response marshaller relies on that nested decoder's presence marker, so it omitted the requested null member.

This deterministic failure currently breaks the package test shard for unrelated pull requests that merge against current `main`.

## Validation

- reproduced `TestBundleIDsResponseMarshalPreservesSparseAttributeFields/null` failing on current `main`
- focused generic-resource and bundle-ID regressions
- `make build`
- `make format`
- `make check-docs`
- `make lint`
- commit hook short suite

No live App Store Connect requests or mutations were performed.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of resources with explicitly null attributes.
  * Ensured null attribute values are preserved correctly when data is processed and saved.

* **Tests**
  * Added coverage for decoding and re-encoding resources with null attributes.

### DIFF
--- a/internal/asc/types/resources.go
+++ b/internal/asc/types/resources.go
@@ -234,7 +234,7 @@ func (r *Resource[T]) UnmarshalJSON(data []byte) error {
 		attributesPresent: decoded.Attributes != nil,
 		attributesNull:    bytes.Equal(bytes.TrimSpace(decoded.Attributes), []byte("null")),
 	}
-	if !r.attributesPresent || r.attributesNull {
+	if !r.attributesPresent {
 		return nil
 	}
 	return json.Unmarshal(decoded.Attributes, &r.Attributes)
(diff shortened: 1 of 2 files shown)


rudrankriyam:
Closing as superseded by merged #2185. That PR replaced the bespoke bundle-ID response marshaler with the generic Response envelope, and the exact null sparse-attribute regression now passes on current main (90812c0e9) without this change. A follow-up audit found no production call site with a remaining observable failure, so keeping this separate generic decoder change would add behavior without a current bug.

Issue #25, opened 2026-01-20
Phase 7: Sandbox tester management


rudrankriyam:
# PRD: Sandbox Testers Management in ASC CLI

## Overview

This PRD outlines the implementation of sandbox tester management for App Store Connect CLI, enabling developers and AI agents to create, list, and delete sandbox testers used for testing in-app purchases and subscriptions.

## Goals

1. Enable developers to create sandbox testers via CLI
2. Support listing and filtering sandbox testers
3. Provide deletion capabilities for sandbox testers
4. Maintain AI-first design with JSON output by default
5. Follow explicit, non-interactive CLI patterns
6. Support clearing purchase history for testers (future enhancement)

---

### Core Sandbox Tester Endpoints

| Method | Endpoint | Description |
|--------|----------|-------------|
| GET | `/v1/sandboxTesters` | List all sandbox testers |
| GET | `/v1/sandboxTesters/{id}` | Get specific sandbox tester details |
| POST | `/v1/sandboxTesters` | Create a new sandbox tester |
| DELETE | `/v1/sandboxTesters/{id}` | Delete a sandbox tester |

### Supporting Endpoints (Future)

| Method | Endpoint | Description |
|--------|----------|-------------|
| POST | `/v1/sandboxTesters/{id}/clearPurchaseHistoryRequest` | Clear purchase history for a tester |

**Note:** Some sources mention `/v2/sandboxTesters`, but `/v1/sandboxTesters` appears to be the standard endpoint. Implementation should verify the correct version during development.

---

### Command Structure

```bash
# List sandbox testers
asc sandbox list [flags]

# Create sandbox tester
asc sandbox create --email "tester@example.com" --first-name "John" --last-name "Doe" --password "SecurePass123!" --territory "USA" [flags]

# Get sandbox tester details
asc sandbox get --id "TESTER_ID" [flags]

# Delete sandbox tester
asc sandbox delete --id "TESTER_ID" [flags]
# OR
asc sandbox delete --email "tester@example.com" [flags]
```

#### `list` - List sandbox testers

Lists all sandbox testers for the team.

**Flags:**
- `--email` (optional) - Filter by email address
- `--territory` (optional) - Filter by territory code (e.g., "USA", "JPN", "GBR")
- `--limit` (optional) - Maximum results per page (1-200)
- `--next` (optional) - Fetch next page using `links.next` URL
- `--output` (optional) - Output format: `json` (default), `table`, `markdown`
[...]

PR #2581, opened 2026-09-15
feat(cli): global read-only mode (ASC_READ_ONLY / --read-only)

### LINKED ISSUE #2563: feat(cli): global read-only mode (ASC_READ_ONLY / --read-only) that refuses all mutations
## Summary
A global read-only mode that refuses every POST/PATCH/PUT/DELETE at the transport layer across all outbound clients (public API, web session, Apple Ads, StoreKit).

## Motivation
Three independent MCP servers built on asc re-implemented "never mutate" (Heimdall `--read-only`, zelentsov asc-mcp read-only, abd3lraouf risk tiers). Agents drive ~48% of invocations; per-command `--confirm` cannot guarantee a session never writes.

## Contract
- `ASC_READ_ONLY=1` env var; a global `--read-only` flag only if the root global-flag mechanism makes it non-invasive.
- Enforced before the request is sent; one stderr line naming the refused method and path; distinct documented exit code.
- GET/HEAD, local-only operations, and JWT minting stay allowed; `--confirm` commands fail fast at validation.

## Acceptance criteria
- [ ] httptest asserting no mutating request leaves any client
- [ ] cmdtest for one public-API mutation and one web mutation (stderr + exit code)
- [ ] GET still works under the mode
- [ ] docs: concepts section, automation guide, docs/design/read-only-mode.md

Partial work: worktree `~/Developer/CLIs/asc-read-only-mode` (branch claude/read-only-mode-20260913).

rudrankriyam:
## Summary

Adds a global read-only mode that refuses every mutating request before it leaves the process. `ASC_READ_ONLY=1` enables it for a session; the root `--read-only` flag enables it for one invocation. Enforcement lives in the transport layer of every outbound client, not in commands, so it also covers the many mutations that never had a `--confirm` flag (`update`, `create`, relationship edits, uploads).

Motivation: three independent MCP servers built on `asc` each re-implemented "never mutate" as a command allowlist that has to be updated whenever `asc` grows a command. This gives them one switch with a transport-level guarantee.

Closes #2563

## Design

- `internal/readonly`: the process-wide switch (env var or root flag; either enables, neither disables the other), `Check(ctx, method, target)` refusing `POST`/`PATCH`/`PUT`/`DELETE`, and `WithReadIntent` for the reads Apple transports as writes.
- Root flag: `--read-only` is bound in `shared.BindRootFlags` next to `--profile`/`--api-debug`, so it is parsed before any subcommand and needs no per-command change. `BindRootFlags` resets it on every bind and `cmd.Run` clears it on return, so one parse cannot leak into the next.
- Refusal: one stderr line, `Error: ASC_READ_ONLY is set; refusing PATCH /v1/builds/123` (or `--read-only is set; ...`), target rendered without its query string so signed upload URLs and filters never reach logs. New exit code `ExitReadOnly = 6`: 2/3/5 already mean usage, Apple auth, and API conflict, and a refusal is a local policy decision with no HTTP status.
- `--confirm` is deliberately *not* a second signal. The issue's contract asked for confirmed commands to fail at validation, and an earlier revision of this branch did that; the review showed it is wrong, because `--confirm` also guards purely local changes (`auth fix`, `auth export-to-config`, `auth remove`, `notarization staple`, keychain operations) that the same contract promises to keep working. The transport check stays the single enforcement point: a confirmed remote mutation is refused when it issues its write (`bundle-ids delete --confirm` sends nothing at all), and any preflight reads it runs first are reads.
- Workflow steps: `asc workflow run` executes each step as a shell command, and the root flag is process-local, so `workflow.buildEnvSlice` forces `ASC_READ_ONLY=1` into the step environment after the step's declared `env`. A nested `asc` call is then refused too. This is inheritance, not containment, and the docs say so: a step's shell command can clear the variable for something it starts, and a step that runs `curl` never reaches these guards.
- Enforcement points: ASC API `newRequest`, upload operations, Notary API plus its S3 single-part and multipart uploads, web session `doRequestBaseWithHTTPClient`, Developer Portal `doDeveloperPortalHTTP`, Apple Ads `requestOnce` and platform-asset upload, StoreKit `request`, ad hoc distribution `S3Store.Ensure`/`ReplaceCorrupt`, Slack notify, `asc snitch` GitHub issue writes, Wall of Apps GitHub client, and the Xcode Cloud usage-alert webhook. Webhook refusals name only the host, because the path carries the secret.
- POST-shaped reads keep working through `readonly.WithReadIntent`: analytics v1/v2 queries, Apple Ads `find`/`query`/reports/search (classified by `EndpointSpec.ReadOnlyRequest`; the raw `asc ads api` passthrough stays method-based), Developer Portal requests carrying `X-HTTP-Method-Override: GET`, team list, agreement history, app-group and website-push-ID listings.
- Deliberately allowed: JWT minting (local), Apple Ads token exchange, web sign-in/2FA/team selection (session establishment), anonymous telemetry (separate opt-out), and local-only commands.

Full write-up: `docs/design/read-only-mode.md`. User docs: `concepts/read-only-mode.mdx` (new), `commands/global-flags.mdx`, `configuration/environment-variables.mdx`, `concepts/error-handling.mdx` (exit code 6), `guides/automation.mdx`, plus regenerated `docs/COMMANDS.md`.

## Evidence

Gates on `e7e934776`, each serialized through the shared host lock:

| Gate | Result |
| --- | --- |
| `make generate-command-docs` | `docs/COMMANDS.md` regenerated and committed (the root flag changed `asc --help`) |
| `make build` | pass |
| `make format` | pass, no diff |
| `make check-docs` | pass (command, repo, website, agent-skill and OpenAPI validators) |
| `make lint` | `0 issues.` |
| `ASC_BYPASS_KEYCHAIN=1 make test` | pass (exit 0); the pre-commit hook re-ran its suite on the commit |

Two failures found and fixed along the way, both worth calling out:

- `internal/cli/docs` `TestASCTemplateIncludesAllRootFlags` failed because a new root flag must also appear in the embedded `asc docs init` template; `internal/cli/docs/templates/ASC.md` now lists `--read-only` and `ASC_READ_ONLY`.
- `make lint` reported `unparam: (*Client).doIdempotentMutation - method always receives "PATCH"`. Diagnosed rather than assumed: linting `internal/asc` with and without the new test, and against a pristine worktree at the branch base, showed the finding only appears with the new test, which had added a third identical call site of that private helper. The assertion was redundant (the loop above it already covers every mutating method through the same `newRequest` choke point) and was removed.

Full-package runs for every touched package (`cmd`, `internal/asc`, `internal/web`, `internal/appleads`, `internal/storekit`, `internal/distribution`, `internal/workflow`, `internal/readonly`, `internal/cli/shared`, `internal/cli/cmdtest`, `internal/cli/web`, `internal/cli/apps`, `internal/cli/notify`, `internal/cli/snitch`, `internal/cli/docs`) pass individually as well.

No live App Store Connect calls were made for this change.

## Test plan

- [x] `internal/readonly`: env parsing and precedence, flag reset, method classification, read intent, message format, query stripping
- [x] `httptest` per client asserting that no mutating request reaches the server while GETs and POST-shaped reads still do: `internal/asc` (API, uploads, notary submissions, notary S3 PUT), `internal/web` (iris, analytics, Developer Portal), `internal/appleads` (selector/report reads allowed, writes refused, passthrough method-based), `internal/storekit`, `internal/distribution` (`Ensure` over httptest, `ReplaceCorrupt` over the existing conditional-replace fake), `internal/cli/notify`, `internal/cli/snitch`, `internal/cli/apps` (Wall of Apps), `internal/cli/web` (usage-alert webhook, asserting the webhook path is not leaked)
- [x] `internal/cli/cmdtest`: stderr line and exit 6 for `builds update` under `ASC_READ_ONLY` and under `--read-only` (naming the flag), `web apps delete --confirm` refused at its PATCH after the preflight GETs, `bundle-ids delete --confirm` refused before any request leaves the transport, `apps list` still succeeding under the mode, and the root flag not leaking after `cmd.Run` returns
- [x] `internal/workflow`: step environments carry `ASC_READ_ONLY=1` exactly once while the mode is on, including when a step's `env` block tries to override it, keep inherited variables, and carry nothing when the mode is off

## Review gate

`codex exec review` with the repository's pinned provider and model, run in the worktree:

1. `--uncommitted` (first round) raised two findings, both valid and both fixed:
   - *P1, workflow subprocesses could bypass the mode.* `asc workflow run` executes steps as shell commands, and the root flag is process-local, so a step invoking `asc` again was unguarded. Fixed by forcing `ASC_READ_ONLY=1` into the step environment (`internal/workflow/env.go`), with a test.
   - *P2, the blanket `--confirm` refusal was wrong.* It rejected commands whose `--confirm` guards only a local change (`auth fix`, `auth export-to-config`, `notarization staple`, keychain operations), contradicting the contract's promise that local-only operations keep working. Fixed by removing the validation-time refusal entirely and leaving the transport as the single enforcement point; `internal/readonly.RefuseCommand` and the shared guard went with it.
2. `--uncommitted` (second round) raised one valid finding: the new docs claimed a workflow file "cannot opt its own steps out", but a step's shell command can still clear the variable for a process it starts. The docs, the design note, and the code comments now describe inheritance rather than containment, and point at a read-only credential for workflow files you do not trust.
3. `--base origin/main` on the final commit `e7e934776`: no findings. Verbatim: "The transport-level guards consistently block mutating requests while preserving explicitly classified POST-shaped reads. The flag lifecycle, exit-code mapping, workflow propagation, documentation, and targeted tests align with the stated read-only contract."

Note on scope: the issue's contract asked for `--confirm` commands to "fail fast at validation". This PR deliberately does not do that, for the reason in finding P2 above; the same contract line promises local-only operations keep working, and the two cannot both hold while `--confirm` also guards local changes. Remote confirmed mutations are still refused before anything leaves the process.


rudrankriyam:
Fixed the failing CodeQL check.

**Alert** #25 `go/allocation-size-overflow` (high) at `internal/workflow/env.go:125`: `forceReadOnlyEnv` preallocated with `make([]string, 0, len(env)+1)`, so the capacity arithmetic was computed from an unbounded entry count.

**Fix** (af77ec13c): the capacity hint now goes through `forcedEnvCapacity`, which clamps the entry count to `maxForcedEnvEntries` (1<<16) before adding the slot for the forced entry, so the addition cannot overflow for any input length. The hint is only a preallocation, so behavior is unchanged: every inherited entry is kept, existing `ASC_READ_ONLY` entries are still stripped, and `ASC_READ_ONLY=1` stays last even past the clamp. Tests: `TestForcedEnvCapacityStaysBounded` (0, 1, clamp-1, clamp, clamp+1, `math.MaxInt-1`, `math.MaxInt`) and `TestForceReadOnlyEnvKeepsEveryEntryBeyondCapacityHint`, which builds an environment larger than the clamp and asserts nothing is dropped or reordered.

**Also fixed from review** (f1a8fdb8f): `EndpointSpec.ReadOnlyRequest` recognized only the `v5` search namespace, so read-only mode refused `POST v1/search/geo` (`platform-resolve-geo-locations`) even though Platform geolocation resolution only reads. The `v1` search prefix is now accepted, with an assertion on the registered endpoint.

**CodeQL result**: alert #25 is now `fixed`, and `refs/pull/2581/merge` has no open code-scanning alerts. CodeQL and every other required check pass on `f1a8fdb8f`.

**Gates on f1a8fdb8f**: `make build`, `make format`, `make check-docs`, `make lint` (0 issues), `ASC_BYPASS_KEYCHAIN=1 make test` all pass locally; full-branch Codex review against `origin/main` reports no actionable findings. Branch is only behind `main`, not conflicting, so it was left unrebased.

Issue #2565, opened 2026-09-14
feat(cli): --if-exists skip|update|fail for the top 409-conflict writes


rudrankriyam:
## Summary
An idempotent-write convention `--if-exists skip|update|fail` (default `fail` = today's behavior) on the writes that generate the most HTTP 409 conflicts.

## Motivation
Telemetry, 30 days, distinct installs hitting 409: pricing availability create 877, review items add 872, metadata push 596, localizations update 562, review submissions-submit 490, review details-create 455, versions create 437, subscriptions setup 424, age-rating edit 409, bundle-ids capabilities add 398. Together the single largest API-error class (~4.7k installs). Agents retry blindly.

## Contract
- `skip`: exit 0 with an additive receipt field (e.g. `action: "skipped"`) via the registered renderer.
- `update`: route to the matching update/PATCH with the same inputs where meaningful (versions create -> versions update; localizations create/update routing; review details-create -> details-update; pricing availability create -> availability edit; capabilities add / review items add -> skip semantics).
- Keyed on Apple's exact 409 error codes; other 409s still fail.

## PR plan
1. shared helper + receipt + `versions create` + `review details-create`
2. `localizations update/create` + `metadata push` (stacked)
3. `pricing availability create` (stacked)
4. `bundle-ids capabilities add` + `review items add` (stacked)
`review submissions-submit`, `subscriptions setup`, `age-rating edit` analyzed and listed as follow-ups.

## Acceptance criteria
- [ ] RED httptest per command reproducing the real 409 body
- [ ] default behavior byte-identical
- [ ] docs/design/if-exists-idempotent-writes.md + command docs regenerated

Partial work: worktree `~/Developer/CLIs/asc-if-exists-core` (branch claude/if-exists-core-20260913).

PR #2583, opened 2026-09-15
feat(cli): --if-exists skip|update|fail for localizations create

rudrankriyam:
## Summary

Second PR in the `--if-exists` series: `asc localizations create --if-exists fail|skip|update`.

`fail` (default) is byte-for-byte today's behavior. `skip` reads the existing localization for that locale back, prints Apple's envelope for it, and exits 0 without touching it. `update` applies the same field flags to the existing localization with `PATCH /v1/appStoreVersionLocalizations/{id}`.

Part of #2565.

## Evidence

Telemetry, 30 days, distinct installs hitting HTTP 409: `localizations update` 562 and `metadata push` 596 sit in the top four of the largest API-error class. The retryable half of that traffic is the duplicate-locale create: an agent creates a locale, the run fails later, the retry re-creates the locale and dies on "already exists" even though the locale is present.

### Exact 409 code keyed on

| Command | Accepted code | Read-back | Evidence |
| --- | --- | --- | --- |
| `localizations create` | `ENTITY_ERROR.ATTRIBUTE.INVALID.DUPLICATE` (pointer `/data/attributes/locale`, detail names the locale and suggests updating) | `GET /v1/appStoreVersions/{id}/appStoreVersionLocalizations`, paginated, matched case-insensitively on `locale` | Detail text reported in Apple developer forum threads 677931 and 776315 ("Entity with locale: ... already exists. Try updating."). Not verified live, since reproducing it is a mutation. `STATE_ERROR.*` (version not editable) keeps failing without a read-back. |

### Scope correction versus the issue's PR plan

The issue listed "`localizations update/create` + `metadata push`" for this PR. Two changes after reading the code:

- **`localizations update` is out of scope, and correctly so.** It is a pure PATCH: it already resolves the localization ID from the locale and, when the locale is absent, fails with its own non-HTTP "no existing localization found" error. Its remaining 409s are state conflicts (version not editable), not existence conflicts, so `--if-exists` has nothing to key on. The idempotency gap the telemetry points at is on the create side, which this PR covers.
- **`metadata push` is deferred to a follow-up.** Every mutation in `push.go` already runs through `shared.RunReconciledMutation` with a field-matching read-back, so the duplicate-locale 409 only survives when the existing remote content differs from the plan. Making that case configurable means threading the mode through `applyVersionLocalizations` and `applyAppInfoLocalizations` and adding a `skipped` status to the `ApplyAction` receipt: a receipt-schema change on a bulk command, which belongs in its own PR rather than bundled here.

## Design

`docs/design/if-exists-idempotent-writes.md` (added in PR1) carries the contract. This PR reuses it unchanged:

- `shared.BindIfExistsFlag` / `shared.ParseIfExistsMode` (usage error, exit 2, before any HTTP request).
- `shared.ResolveIfExistsConflict` requires HTTP 409 **and** the exact Apple code **and** a successful natural-key read-back.
- `localizations create` prints Apple's envelope unmodified in all three modes, per the JSON output contract; `skip` and `update` write one stderr diagnostic naming the existing localization ID.
- `client.UpdateAppStoreVersionLocalization` already strips the immutable `locale` from the PATCH body, and `AppStoreVersionLocalizationAttributes` is `omitempty` throughout, so `update` sends exactly the fields the caller passed.

## Test plan

- [x] RED-first: the new `httptest` cases in `internal/cli/cmdtest/if_exists_localizations_test.go` were run against this branch's base with `internal/cli/localizations/create.go` reverted and failed (`exit status 2`, the flag does not exist), then passed with the implementation in place.
- [x] Coverage: `skip` (POST, locale read-back, detail re-read, Apple's envelope, stderr line, exit 0), `update` (PATCH carries the supplied fields and never the immutable `locale`), `update` with nothing updatable (no PATCH at all), `fail` (only the POST, Apple's detail preserved, empty stdout), a 409 `STATE_ERROR` (fails with no read-back), a matching 409 whose read-back finds nothing (still fails), an invalid `--if-exists` value (usage error before any HTTP request), and no submit-readiness "was created" warning for a handled duplicate.
- [x] `go test ./internal/cli/localizations/` and the whole `LocalizationsCreate` set in `./internal/cli/cmdtest/` pass, including the 17 pre-existing `localizations create` warning tests, so the readiness-warning change is not a regression.
All run on the pushed head `a2bc998e8` via the shared gate lock, each exit code captured directly. `docs/COMMANDS.md` needed no regeneration: it is generated from the top-level `asc --help` and does not list per-subcommand flags.

| Gate | Result |
| --- | --- |

No known-flaky test fired in this run.

## Stacking note

PR2 of 4, based on `claude/if-exists-core-20260913` (PR1). **The maintainer must re-base this branch onto `main` after PR1 merges**; the diff shown against PR1's branch is this PR's own change only.

## Review gate

**Pre-commit run** (`--uncommitted`): two findings, both valid, both fixed before committing.

1. [P2] The submit-readiness *create* warning still ran for a handled duplicate, so a locale that already carried the omitted fields was reported as created without them. Fixed: the warning and its version lookup now happen only when the create actually succeeded, and `TestLocalizationsCreateIfExistsSkipEmitsNoCreateReadinessWarning` pins it.
2. [P2] `skip` printed an envelope synthesized from the collection read-back, dropping Apple's top-level `links`/`included`/`meta` and violating the JSON output contract in `AGENTS.md`. Fixed: the resolved localization is re-read with `GET /v1/appStoreVersionLocalizations/{id}` so Apple's own single-resource response is printed unmodified. Re-running `--uncommitted` after the fixes came back clean.

**Post-commit full-branch run** (`--base claude/if-exists-core-20260913`): one finding, valid, fixed in `a2bc998e8`.

3. [P2] A locale-only create resolved with `--if-exists update` sent `"attributes":{}` and reported the localization as updated. That is an unnecessary write a non-editable localization can reject. Fixed: with no metadata field supplied, `update` now resolves like `skip`, matching `versions create --if-exists update`, which already skips its PATCH when neither `--copyright` nor `--release-type` is set. `TestLocalizationsCreateIfExistsUpdateWithoutUpdatableFieldsDoesNotPatch` pins it, and the design doc records the rule for the whole convention.

**Final full-branch run** on the committed head `a2bc998e8` (`--base claude/if-exists-core-20260913`): **clean**, no actionable findings.

No finding was dispositioned as a false positive in this PR.


rudrankriyam:

**Merge:** `d9b72aefd` — `Merge claude/if-exists-core-20260913 into claude/if-exists-localizations-20260915`

**Conflict resolved:** `docs/design/if-exists-idempotent-writes.md`, in the
"Error codes keyed on, per command" table. Core rewrote the `versions create`
and `review details-create` rows with live-verified evidence while this branch
rewrote the adjacent `localizations create` row, so the three-row block
conflicted. Resolution keeps core's two rows verbatim and this branch's
`localizations create` row (which also narrows the row to `localizations
create`, since this branch moves `metadata push` to a follow-up).

Everything else in that doc auto-merged and both sides' prose is intact:
this branch's "empty PATCH resolves like `skip`" paragraph, the
submit-readiness-advice suppression bullet, and the rollout-item-2 rewrite;
plus core's rollout-item-1 note about `--copy-metadata-from` reporting
`updated`. No conflict markers remain.

`internal/cli/shared/if_exists.go` and its test auto-merged to core's new
version, so this branch now also rejects an explicitly empty `--if-exists ""`.

**Gates on head `d9b72aefd`** (serialized through the shared host lock):

- `make build` — pass
- `make lint` — 0 issues
- focused: `go test ./internal/cli/localizations/... ./internal/cli/shared/... ./internal/cli/cmdtest/ -run 'IfExists|Localization'` — pass

PR #2579, opened 2026-09-15
feat(cli): --if-exists skip|update|fail for versions create and review details-create

rudrankriyam:
## Summary

Adds the shared `--if-exists fail|skip|update` convention for idempotent create-style writes and wires it into the first two commands:

- `asc versions create --if-exists skip|update`
- `asc review details-create --if-exists skip|update`

`fail` (default) is byte-for-byte today's behavior. `skip` treats an existing resource as success (exit 0, resource untouched, one stderr diagnostic). `update` routes the same inputs to the resource's PATCH (`versions update` with `--copyright`/`--release-type`; `review details-update` with the same attributes). Unknown values are usage errors (exit 2) before any HTTP request.

Closes #2565.

## Evidence

Telemetry, 30 days, distinct installs hitting HTTP 409: `pricing availability create` 877, `review items add` 872, `metadata push` 596, `localizations update` 562, `review submissions-submit` 490, `review details-create` 455, `versions create` 437, `subscriptions setup` 424, `age-rating edit` 409, `bundle-ids capabilities add` 398. This is the single largest API-error class (~4.7k installs); agents retry blindly and fail on "already exists" even though the desired end state is present.

### Exact 409 codes keyed on

The helper (`shared.ResolveIfExistsConflict`) only handles an HTTP 409 whose `errors[0].code` is on the command's recorded list **and** whose natural-key read-back returns the resource. Anything else (including `STATE_ERROR.*`) fails exactly as before; unlisted codes never trigger the read-back.

| Command | Accepted code(s) | Read-back | Evidence |
| --- | --- | --- | --- |
| `versions create` | `ENTITY_ERROR.ATTRIBUTE.INVALID.DUPLICATE` (pointer `/data/attributes/versionString`, "The version number has been previously used.") | `GET /v1/apps/{id}/appStoreVersions?filter[versionString]&filter[platform]` | Apple body quoted in developer forum thread 741290 and fastlane/codemagic reports. The same endpoint returns 409 `ENTITY_ERROR.RELATIONSHIP.INVALID` ("You cannot create a new version of the App in the current state."), which is covered by a test and keeps failing. |
[...]

PR #2584, opened 2026-09-15
feat(cli): --if-exists skip|update|fail for pricing availability create

rudrankriyam:
## Summary

Third PR in the `--if-exists` series: `asc pricing availability create --if-exists fail|skip|update`.

`fail` (default) is byte-for-byte today's behavior. `skip` reads the existing `appAvailability` back, prints it, and exits 0 without touching it. `update` applies `--territory` and `--available` to the existing record through the same code path as `asc pricing availability edit`.

Part of #2565.

## Evidence

Telemetry, 30 days: `pricing availability create` is the **largest single source of HTTP 409s in the CLI** at 877 distinct installs. Availability bootstrap is exactly the write an agent retries after a partial setup run, and the retry dies on "already exists" even though the record is present.

### Exact 409 codes keyed on

| Command | Accepted codes | Read-back | Evidence |
| --- | --- | --- | --- |
| `pricing availability create` | `ENTITY_ERROR.RELATIONSHIP.INVALID` (pointer `/data/relationships/app`), `ENTITY_ERROR.ATTRIBUTE.INVALID.ALREADY_EXISTS` | `GET /v1/apps/{id}/appAvailabilityV2` | `internal/asc/pricing_test.go` already replays a 409 `ENTITY_ERROR.RELATIONSHIP.INVALID` for `POST /v2/appAvailabilities`, and `web apps availability create` already treats an existing record as this same condition. Codes not verified live (reproducing the duplicate is a mutation); the read-back is the decisive check. |

### The ambiguous code, handled explicitly

Apple answers **both** the existence conflict **and** its public-API bootstrap rejection with 409 `ENTITY_ERROR.RELATIONSHIP.INVALID` on this endpoint. The command therefore keeps classifying the bootstrap rejection **first**, by its distinctive detail (`territoryAvailabilities.territory` / "expects an included resource with type 'territories'"), so that case keeps its existing web-session remediation and is never swallowed by `--if-exists`. That ordering is pinned by a test. The read-back is a second independent guard: the bootstrap rejection creates nothing, so a read-back would find nothing regardless.

`STATE_ERROR.*` and a matching 409 whose read-back finds no record keep failing, each with its own test.

### `availableInNewTerritories` on `update`

Apple exposes no update operation for `availableInNewTerritories`. On `update` the flag is therefore only **verified** against the existing policy and a mismatch fails — which is exactly what `pricing availability edit` already does with `--available-in-new-territories`. `--territory` and `--available` are applied. This is documented in the command help and in the design doc.

## Design

`docs/design/if-exists-idempotent-writes.md`, updated by this PR with the pricing row and the ambiguity note.

The `update` route goes through a new exported `shared.ApplyTerritoryAvailabilityUpdate` wrapping the existing `executeTerritoryAvailabilityUpdate` implementation, so the edit logic (policy verification, territory diffing, PATCH batching, post-update verification) is reused rather than duplicated. `skip` still pays for the territory-catalog fetch the create performs before its POST; the help text says so.

## Test plan

- [x] RED-first: the new `httptest` cases in `internal/cli/cmdtest/if_exists_pricing_test.go` were run with `internal/cli/pricing/pricing.go` and `internal/cli/shared/availability_command.go` reverted to the base and failed (`exit status 2`, the flag does not exist), then passed with the implementation in place.
- [x] Coverage: `skip` (territories, POST, read-back, Apple's envelope, stderr line, exit 0), `update` (routes into the edit path — asserted by its second availability read and its own territory-availabilities fetch — and issues no PATCH when both territories already match), `fail` (territories and POST only, Apple's detail preserved, empty stdout), the bootstrap rejection (keeps its own remediation, no read-back), `STATE_ERROR` (no read-back), a matching 409 whose read-back 404s (still fails), and an invalid `--if-exists` value (usage error before any HTTP request).
- [x] `go test ./internal/cli/pricing/ ./internal/cli/shared/` pass.
All run on the pushed head `c9e116b2f` via the shared gate lock, each exit code captured directly.

| Gate | Result |
| --- | --- |

`docs/COMMANDS.md` needed no regeneration: it is generated from the top-level `asc --help` and does not list per-subcommand flags. No known-flaky test fired.

## Stacking note

PR3 of 4, based on `claude/if-exists-core-20260913` (PR1). **The maintainer must re-base this branch onto `main` after PR1 merges**; the diff against PR1's branch is this PR's own change only. It does not depend on PR2.

## Review gate

**Pre-commit run** (`--uncommitted`): **clean**, no actionable findings.

**Lint then caught what codex did not**: `make lint` in the pre-commit hook failed with staticcheck `S1016` on the new wrapper (it built `availabilityUpdateRequest` with a struct literal where the two identical structs permit a conversion). Fixed by using `availabilityUpdateRequest(request)`, verified with `golangci-lint run --concurrency=2 ./internal/cli/shared/...` reporting `0 issues.`, and `--uncommitted` was re-run afterwards and came back clean again.

**Post-commit full-branch run** on `c9e116b2f` (`--base claude/if-exists-core-20260913`): **clean**, no actionable findings. Codex's summary: the conflict classification, read-back requirement, skip behavior, and update routing align with the documented idempotency contract.

No finding was dispositioned as a false positive in this PR.


rudrankriyam:

**Merge:** `2257dd2e3` — `Merge claude/if-exists-core-20260913 into claude/if-exists-pricing-20260915`

**Conflicts:** none. `docs/design/if-exists-idempotent-writes.md` and
`internal/cli/pricing/pricing.go` auto-merged.

This picks up two things from core: the `main` restack (#2579 merged current
`main`, resolving the `internal/asc/errors_test.go` conflict with #2574) and a
core review fix that makes an explicitly empty `--if-exists ""` a usage error
instead of silently meaning `fail`. `pricing availability create` inherits that
through `shared.ParseIfExistsMode`; no change was needed on this branch.

**Gates on head `2257dd2e3`** (serialized through the shared host lock):

- focused: `go test ./internal/cli/pricing/... ./internal/cli/iap/... ./internal/cli/shared/... ./internal/cli/cmdtest/ -run 'IfExists|Availability'` — pass
- `make build` — pass
- `make lint` — 0 issues

PR #2586, opened 2026-09-15
feat(cli): --if-exists for bundle-ids capabilities add and review items-add

rudrankriyam:
## Summary

Fourth and last PR in the `--if-exists` series:

- `asc bundle-ids capabilities add --if-exists fail|skip|update`
- `asc review items-add --if-exists fail|skip` (`update` is a usage error)

`fail` (default) is byte-for-byte today's behavior in both commands.

Part of #2565.

## Evidence

Telemetry, 30 days, distinct installs hitting HTTP 409: `review items add` 872 (second largest in the CLI) and `bundle-ids capabilities add` 398. Both are pure attach/enable writes, so the desired end state after an "already exists" 409 is already present and the retry should succeed.

### Exact 409 codes keyed on

| Command | Accepted codes | Read-back | Evidence |
| --- | --- | --- | --- |
| `bundle-ids capabilities add` | `ENTITY_ERROR.ATTRIBUTE.INVALID.DUPLICATE`, `ENTITY_ERROR.ATTRIBUTE.INVALID.ALREADY_EXISTS` | `GET /v1/bundleIds/{id}/bundleIdCapabilities`, paginated, matched case-insensitively on `capabilityType` | Codes not verified live (reproducing the duplicate is a mutation); the read-back is the decisive check. |
| `review items-add` | `ENTITY_ERROR.RELATIONSHIP.INVALID`, `ENTITY_ERROR.ATTRIBUTE.INVALID.ALREADY_EXISTS`, `ENTITY_ERROR.ATTRIBUTE.INVALID.DUPLICATE` | `GET /v1/reviewSubmissions/{id}/items?include=<relationship>`, paginated, matched on the linked resource ID | Codes not verified live; the read-back is the decisive check. |

Kept failing, each with its own test: `ENTITY_ERROR.ATTRIBUTE.TYPE` (a capability type Apple will not accept for the bundle ID — also a 409) and `STATE_ERROR.*` (a review submission that is no longer editable).

### Three read-back subtleties this had to get right

1. **`include=`, not `fields[]`.** With `fields[]` alone Apple returns submission items carrying `links` only and no `relationships` key at all, so every item would read as an empty ID and the read-back would never match. `internal/cli/submit/submit_create.go` already documents this, and the OpenAPI snapshot confirms `include` accepts exactly these relationship names. Pinned by asserting `include=appStoreVersion` on the read-back request.
2. **A relationship can be `"data":null`.** Apple includes relationship keys for types an item does not carry, so a non-nil pointer is not a match — a real ID must be equal. `submissions_history.go` documents the same trap.
3. **`REMOVED` items are not proof of presence.** A matching item whose `state` is `REMOVED` is historical: the resource is detached from the submission. Taking it as proof would make `skip` report success while the item is still not attached. Skipped, matching `background_assets_submit.go`, which applies the same exclusion when deciding which versions are already attached.

### Why the envelope is built rather than re-read

Apple exposes **no detail endpoint** for either resource. The OpenAPI snapshot has only `POST`, `PATCH` and `DELETE` for `/v1/bundleIdCapabilities/{id}` and `/v1/reviewSubmissionItems/{id}`; there is no `GET`. The collection item is therefore the only representation Apple offers, so the printed single-resource envelope is built around Apple's own resource object verbatim. Where a detail endpoint does exist, this convention re-reads instead (PR2's `localizations create`, PR1's `review details-create`). The design doc records the distinction.

### Why `review items-add` has no `update`

A submission item carries no inputs to re-apply — `review items-update` only sets `resolved`/`removed`, which are not create inputs. So the command binds only `skip`, and `--if-exists update` is a usage error (exit 2) whose message lists just `fail, skip`. Pinned by a test that also asserts `update` is never advertised.

## Design

`docs/design/if-exists-idempotent-writes.md`, updated by this PR with both rows, the `include=`/`null`/`REMOVED` notes, and the no-detail-endpoint rationale.

`bundle-ids capabilities add --if-exists update` routes `--settings` to `PATCH /v1/bundleIdCapabilities/{id}`. With no `--settings` there is nothing to apply, so it resolves like `skip` rather than sending an empty PATCH — the same rule PR1's `versions create` and PR2's `localizations create` follow.

## Test plan

- [x] RED-first: the new `httptest` cases in `internal/cli/cmdtest/if_exists_capabilities_test.go` were run with both command files reverted to the base and failed (`exit status 2`, the flags do not exist), then passed with the implementation in place.
- [x] 12 cases covering, per command: `skip`, `update` (capabilities: PATCH carries the settings and never the create-only `capabilityType`), `update` with nothing to apply (no PATCH), `fail` (Apple's detail preserved, empty stdout, no read-back), the non-existence 409s, a read-back that finds nothing, a read-back that finds only a different linked resource, a read-back that finds only a `REMOVED` item, and `review items-add --if-exists update` rejected as a usage error.
- [x] `go test ./internal/cli/bundleids/ ./internal/cli/reviews/` pass.
All run on the pushed head `9d8f562ff` via the shared gate lock, each exit code captured directly.

| Gate | Result |
| --- | --- |

`docs/COMMANDS.md` needed no regeneration: it is generated from the top-level `asc --help` and does not list per-subcommand flags. No known-flaky test fired.

## Stacking note

PR4 of 4, based on `claude/if-exists-core-20260913` (PR1). **The maintainer must re-base this branch onto `main` after PR1 merges**; the diff against PR1's branch is this PR's own change only. It does not depend on PR2 or PR3.

## Review gate

**Pre-commit run** (`--uncommitted`): two findings.

1. [P2] "Ignore removed review items during conflict lookup" — **valid, fixed before committing.** A matching item whose `state` is `REMOVED` is detached from the submission, so returning it would make `skip` report success while the item is still not attached. The read-back now skips `REMOVED` items, matching `background_assets_submit.go`, and `TestReviewItemsAddIfExistsSkipIgnoresRemovedItems` pins it. Re-running `--uncommitted` afterwards raised finding 2 in its place.
2. [P1] "Regenerate command docs after changing help" — **false positive, verified.** `docs/COMMANDS.md` is generated from the top-level `asc --help` (`scripts/generate-command-docs.py` runs `go run . --help`), so it lists command groups and subcommand names and never per-subcommand flags. The repository's own gate agrees on this exact head: the pre-commit hook's `generate-command-docs.py --check` and `check-commands-docs.py` both passed (the commit could not have completed otherwise, since the hook runs under `set -euo pipefail`), and `make check-docs` exited 0 in the gate run above. `make generate-command-docs` produces no diff. No change made. Codex raised the identical claim on PR1 and it was disproven the same way.

**Post-commit full-branch run** on `9d8f562ff` (`--base claude/if-exists-core-20260913`): one finding.

3. [P2] "Honor explicitly empty capability settings" (`--settings '[]'` should clear settings) — **not actionable in this PR; recorded as a follow-up above.** The `len(settingsValue) > 0` check deliberately mirrors the stable `bundle-ids capabilities update` contract, whose code says "Treat empty settings arrays as no-op updates" and which rejects `--settings '[]'` with no `--capability` as a usage error. Additionally `Settings` carries `json:"settings,omitempty"` on all three shared capability structs, so `[]` cannot be serialized without changing stable API surface shared with that command. Making `[]` mean "clear" is a behavior change on stable surface that needs the deprecation ladder and live confirmation Apple accepts `"settings":[]`. Note the current behavior makes no false claim: with `--settings '[]'` the command reports `left unchanged`, which is accurate.

Findings 1 was fixed; 2 was disproven; 3 is dispositioned as out-of-scope with a follow-up. No further code change was made after the reviewed head, so the gate results above apply to the exact PR head.

## Follow-ups this PR deliberately does not take

- **`--settings '[]'` as an explicit "clear settings".** Codex raised it (below). Out of scope here: `bundle-ids capabilities update` already defines `--settings '[]'` as a no-op (its code says so, and `--settings '[]'` with no `--capability` is a usage error there), and `Settings` carries `json:"settings,omitempty"` on all three shared capability structs, so `[]` cannot be serialized without changing stable API surface that the existing `capabilities update` command shares. Making `[]` mean "clear" is a behavior change to a stable command, needs the deprecation ladder, and needs live confirmation that Apple accepts `"settings":[]` at all.
- `metadata push` idempotency (deferred from PR2), plus `review submissions-submit`, `subscriptions setup` and `age-rating edit`, which the design doc analyzes and rules out as existence conflicts.


rudrankriyam:

**Merge:** `24d999a` — `Merge claude/if-exists-core-20260913 into claude/if-exists-capabilities-20260915`

**Conflicts:** none. `docs/design/if-exists-idempotent-writes.md` auto-merged.

This picks up two things from core: the `main` restack (#2579 merged current
`main`, resolving the `internal/asc/errors_test.go` conflict with #2574) and a
core review fix that makes an explicitly empty `--if-exists ""` a usage error
instead of silently meaning `fail`. `bundle-ids capabilities add` and `review
items add` inherit that through `shared.ParseIfExistsMode`; no change was needed
on this branch.

**Gates on head `24d999a`** (serialized through the shared host lock):

- focused: `go test ./internal/cli/bundleids/... ./internal/cli/reviews/... ./internal/cli/shared/... ./internal/cli/cmdtest/ -run 'IfExists|Capabilit|ItemsAdd'` — pass
- `make build` — pass
- `make lint` — 0 issues

PR #2618, opened 2026-09-15
feat(cli): name stray positional operands and the flag they probably meant

### LINKED ISSUE #2613: feat(cli): name stray positional operands and the flag they probably meant
## Summary
453 distinct installs across 251 commands (5.x, 14 days) passed a stray positional operand to a command that takes only flags, e.g. `asc apps view 123`, `asc pricing availability create 37`, `asc builds list 33`, `asc auth login 26`, `asc bundle-ids list 26`, `asc screenshots list 23`. The current error does not say which token was unexpected or what to do.

## What to do
In the shared parse path, when a leaf command receives unexpected positionals, print `Error: unexpected argument "123"` plus, when the command has exactly one required or primary ID flag (`--id`, `--app`, `--build-id`, …), `Did you mean: asc apps view --id 123`. Exit 2. Never echo values into telemetry.

## Acceptance criteria
- [ ] cmdtest for five commands above
- [ ] commands that legitimately take positionals (`search`, `schema`, `api`, `docs`) unaffected

rudrankriyam:
## Summary

453 distinct installs across 251 commands passed a bare operand to a command that takes only flags. Until now the CLI either dropped the token (`asc builds list 33` ran a full listing) or reported only the missing flag (`asc apps view 123` said `--id is required`). The shared parse path now rejects the operand before the command runs, names the token, and points at the flag it belongs to.

```
$ asc apps view 123
Error: unexpected argument "123"
Did you mean: asc apps view --id 123
For help:
  asc apps view --help
```

Exit code is 2, the diagnostic goes to stderr, and nothing is written to stdout.

## How the hint is chosen

The primary identifier flag comes from the command's own flag set, matched against a small allowlist: `--id`, `--app`, `--build-id`, `--version-id`, `--group-id`, `--iap-id`, `--subscription-id`, `--localization-id`. The hint is printed only when the command defines exactly one of them and the caller has not already passed it, so it never guesses:

- `asc screenshots list 23` defines both `--app` and `--version-id` -> operand named, no hint.
- `asc auth login 26` defines none of them -> operand named, no hint.
- `asc iap list --app app-1 unexpected` already has `--app` -> operand named, no hint.
- Several stray operands -> all of them named (`unexpected arguments "123", "456"`), no hint.

When flags follow the stray token they were never parsed, so only the operands up to the first flag spelling are named. That keeps `asc apps search-keywords set unexpected --app app-1 ...` from burying the one bad token under the eight flags the caller got right.

## Exclusions

Commands whose operands are part of their contract are untouched: `asc search`, `asc schema`, `asc docs show`, `asc workflow run`, `asc signing run` (its `--` passthrough), and the `asc snitch` group, along with every other command group, which keeps reporting its own unknown-child diagnostics. A `--` terminator is not itself an exemption: it stops flag parsing, but it does not make an operand usable by a flag-only command, so `asc apps view --id APP_ID -- stray` is rejected too.

The hint is rendered through the same helpers as the existing command-path recovery suggestions: POSIX shells get the operand single-quoted, and on Windows an operand that cannot be rendered safely for both cmd.exe and PowerShell drops the hint and keeps the named token.

## Telemetry

The rejected token is unbounded caller-supplied text and never reaches telemetry. `FailureParameter` stays empty and the event is a plain usage failure for the command path. A test builds the event and asserts the serialized payload does not contain the operand.

## Tests

- `internal/cli/cmdtest/stray_positional_args_test.go`: the five commands from the issue plus the no-hint, multi-operand and terminator cases, and the exclusions still succeeding.
- `cmd/stray_positionals_test.go`: the telemetry privacy contract, per-shell hint rendering, a tree walk over all 1427 leaf commands asserting the exclusion list and the commands' own declared usage agree in both directions, and the group carve-out.
- `internal/cli/shared/stray_positionals_test.go`: the flag allowlist and usage-declaration parsing.

Existing per-command positional diagnostics kept their own wording in-package; the tests that drove them through the root now assert the unified message.

Closes #2613


rudrankriyam (commented):
Blocking audit finding before merge:

`cmd/stray_positionals.go` renders the copyable POSIX correction with `shellSafeCommandArg`, which first calls `SanitizeTerminal`. For control characters, ANSI sequences, bidi marks, or invalid UTF-8, the printed command can therefore contain a different operand from the one the caller supplied (for example, `line\nnext` becomes `line next`). A copyable hint must be byte-exact or omitted. Please route this path through the stricter `shared.ShellQuote` contract (or an OS-parameterized equivalent), omit the hint when rendering is unsafe, and add newline/ANSI/bidi/invalid-UTF8 regression tests. The named error can remain sanitized.

PR #2621, opened 2026-09-16
feat(web): manage Sign in with Apple private keys

alexdotpink:
Adds `asc web sign-in-keys list`, `view`, `create`, and `download` so a Sign in with Apple private key can be configured for a primary Bundle ID and downloaded without opening Developer Portal. Existing App Store Connect API-key commands are unchanged.

The commands reuse Developer Portal sessions and team selection, verify the created key bundle association, and preserve the Apple JSON envelope for reads. Create and standalone download require explicit `--confirm`; create/download prepare private recovery files before contacting Apple, never overwrite an existing key, and return a receipt without exposing key material. One-time operations are not automatically retried. Private file publication requires macOS or Linux.

Validation includes HTTP payload/header checks, command validation and receipts, destination collision/replacement coverage, inherited macOS ACL protection, and Windows fail-before-mutation behavior. Live list, view, create, and P8 download were verified by the contributor with a dedicated app key; existing keys were preserved. These commands use private Developer Portal endpoints, which may change.

Validation on the final local head: `make format`, `make build`, `make generate-command-docs`, `make check-docs`, `make lint`, and `ASC_BYPASS_KEYCHAIN=1 make test` passed. Black-box and focused checks confirmed both irreversible paths require `--confirm` before session resolution or output-directory creation. Isolated pre-commit reviews reported no actionable findings; the exact committed branch is reviewed again after every diff change.


rudrankriyam:
Thank you!

PR #2623, opened 2026-09-16
fix(optimize): omit popularity sort on Ads keyword suggestions

### LINKED ISSUE #2622: optimize keywords discover fails: suggestions API rejects sorting by popularity
### asc version

5.2.1 (commit 5321681)

### OS

macOS (darwin/amd64)

### What happened

`asc optimize keywords discover` always fails against the live Apple Ads Platform API. Auth is fine — `asc ads auth status --validate` passes, and `asc optimize keywords score --genre PHOTO_VIDEO` still returns search-term popularity. Only discover blows up.

It looks like discover always sends:

```json
"sorting": [{ "field": "popularity", "order": "DESC" }]
```

Apple rejects that on `v1/suggestions/keywords/query`:

```
HTTP 400 VALIDATION_ERROR
INVALID_FIELD_ATTRIBUTE sorting[0].popularity
Field 'popularity' is not queryable
```

At the same time `v1/suggestions/phrases/query` keeps returning HTTP 500 (retried a few times, same result). Discover then fails hard because both suggestion sources are unavailable.

This feels related to the sort-field mismatch in #2038, but that one was about `search-term-popularity`. Here it's the suggestions endpoints, and popularity simply isn't a sortable field anymore.

### Reproduction

With a working Ads profile / ad account:

```bash
asc ads auth status --validate

asc optimize keywords discover \
  --app YOUR_APP_ID \
  --country US \
  --limit 10 \
  --ad-account YOUR_AD_ACCOUNT_ID
```

### Expected

Discover should return keyword (and ideally phrase) suggestions, or at least degrade gracefully if one of the two sources is down.

### Workaround / fix that worked for me

I checked `internal/cli/ads/search_optimization.go` in `fetchOptimizationSuggestionsLimitedWithMore`. The request body hardcodes the popularity sort. Apple's own starter payload for suggestions also doesn't include sorting.

Locally I just dropped the `sorting` field and left filters as-is. After that:

- `keyword_suggestions` came back as `available`
- `phrase_suggestions` still 500 from Apple (separate issue)
- `asc optimize keywords discover` exited successfully using the keyword source

Client-side sorting by whatever `popularity` comes back on each row still works fine in `keywords_discover.go`, so omitting the request sort doesn't hurt much.

Happy to open a PR if useful — change is basically:

```go
body := map[string]any{
  "filters": filters,
  // omit sorting — Apple rejects field "popularity" on suggestions
}
```

plus updating `TestFetchSearchSuggestionsSortsByPopularityAndHonorsLimit` so it asserts sorting is absent instead of required.

### Notes

Even after the sort fix, the keyword suggestions I got looked oddly generic (top-of-store type terms) rather than app-specific. Not sure if that's expected for this endpoint without a campaign context, or another filtering issue. Separate from the 400 though.

fxmumu:
## Summary
- `asc optimize keywords discover` currently sends `sorting: [{ field: popularity }]` to Apple's suggestions endpoints, which reject that field with `INVALID_FIELD_ATTRIBUTE`.
- This PR omits request-side sorting (matches Apple's starter payload). Results are still ordered locally by returned popularity when present.
- Updates the unit test to assert sorting is omitted.

Fixes #2622.

## Test plan
- [x] `go test ./internal/cli/ads/ -run TestFetchSearchSuggestionsOmitsSortingAndHonorsLimit`
- [x] Live check: `asc optimize keywords discover --app … --country US` — `keyword_suggestions` becomes `available` (phrase endpoint may still 500 on Apple's side)


fxmumu:
One follow-up while this is under review — separate from the 400 fix itself.

Omitting the unsupported `sorting` field still looks like the right fix to unblock `discover`. The local popularity re-sort in `buildKeywordDiscoverReport` was already there before this PR.

After the endpoint works again though, that re-sort may be a weaker stand-in for server-side popularity ordering than the PR description suggested: popularity isn't the same as app relevance, and without request-side sort the first page is whatever Apple's default order is. Re-ranking that page by popularity can push generic high-volume terms to the top (same vibe as the "oddly generic" note on #2622).

Fine keeping this PR scoped to "omit invalid sorting" only. If useful I can follow up with either:
1. preserve Apple's response order by default, or
2. an explicit `--sort popularity|api` flag

Prefer follow-up PR, or fold a minimal order-preserving change into this one?

PR #2636, opened 2026-09-16
Append telemetry records instead of rewriting the spool

rudrankriyam:
## Summary
- Each enabled telemetry command was locking the spool, decoding every queued record, rewriting the file, and calling `F_FULLFSYNC` twice before exit. Append now extends the existing file under the same lock. The worker snapshot still trims to 128 records and repairs a truncated line.
- A 3xx from the collector is now a permanent delivery error. Redirects are not followed, so retrying them only pinned the spool at its maximum rewrite cost. 5xx and transport failures stay retryable.

## Evidence
- `TestSpoolAppendExtendsTheExistingFile` fails on the old rewrite path because the inode changes, and passes when the second record is appended to the same file.
- `TestSendHTTPEventClassifiesPermanentCollectorRejections` now covers 301, 302, 307, and 308.
- `ASC_BYPASS_KEYCHAIN=1 go test ./internal/telemetry -count=1` passed before commit. The pre-commit short suite also passed.

## Test plan
- [x] Focused telemetry package tests
- [ ] GitHub checks


rudrankriyam:
Closing this before merge. The follow-up audit says telemetry cost claims were not measured, and performance work should be benchmarked before it is optimized. The append-path change can be reopened from `codex/telemetry-fast-append` if a benchmark shows the spool rewrite is actually on the critical path.

